PR #18, opened 2022-02-02
Fix bug with import mask for positive class

darshats:
Also removed the offensive print


choosehappy:
Thanks so much! Apologies as well for the offensive print statement, completely slipped through my code review, not that I can read mandarin anyway :-\

PR #37, opened 2023-08-23
added touchscreen support

jacksonjacobs1:
Added event listeners for touchscreen support.

Touchscreen functionality was tested with the following devices:

- iPad
- Dell precision 7670 touch screen
- Wacom intuos drawing pad

### Issues:
QuickAnnotator UI was not developed responsively, thus smaller displays such as iPads and mobile devices may not display correctly. However, support for these devices might be outside of the scope of this project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+# Ignore specific files
+QA.log
+data.db
+data.db-shm
+data.db-wal
+flask.log
+jobs.log
+
+# Ignore a directory and its contents
+projects/
+
+# Ignore Python bytecode cache directories
+__pycache__/
--- a/templates/annotation-tool.js
+++ b/templates/annotation-tool.js
@@ -522,7 +522,76 @@ function getSquaredDistance(x1, y1, x2, y2) {
 }
 
 function prepareToolMouseListeners(which_canvas){
+    
+    ////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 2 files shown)


choosehappy:
Don't want to be a naysayer here, but these code changes look very similar to the "window.addEventListener('mousemove', function(event) {" function?

is there a specific requirement as to why these need to be separated, or can these be refactored to use the same set of code? 

i mostly worry because this stuff is pretty brittle, and it is unlikely that if we make changes to one section we'll remember to make the same changes to other --- yielding different results from the same actions

thoughts?

jacksonjacobs1:
That makes sense! I can move the bulk of the code into a helper function which is shared by the touch and mouse event listeners. Would that make sense to you?

choosehappy:
yup
________________________________
From: Jackson Jacobs ***@***.***>
Sent: Thursday, August 24, 2023 3:21 PM
To: choosehappy/QuickAnnotator ***@***.***>
Cc: Janowczyk, Andrew R. ***@***.***>; Comment ***@***.***>
Subject: [External] Re: [choosehappy/QuickAnnotator] added touchscreen support (PR #37)


That makes sense! I can move the bulk of the code into a helper function which is shared by the touch and mouse event listeners. Would that make sense to you?

choosehappy:
yum!

PR #49, opened 2024-12-16
Add ray to QuickAnnotator

jacksonjacobs1:
The following changes were made in this PR:
1. Added CLI option for connecting to an existing ray cluster
2. Implemented async prediction generation using the ray cluster. 
3. Updated readme


quickannotator/api/v1/tile/helper.py
```diff
@@ -76,35 +79,62 @@ def generate_random_circle_within_bbox(bbox: Polygon, radius: float) -> shapely.
...
+            # Process the tile (using shapely for example)
             bbox = shapely.wkb.loads(tile.geom.data)
-            for i in range(random.randint(20, 40)):
+            for _ in range(random.randint(20, 40)):
                 polygon = generate_random_circle_within_bbox(bbox, 100)
                 insert_new_annotation(session, image_id, annotation_class_id, is_gt=False, polygon=polygon)
 
-            # now that the tile has been processed, update the tile.seen column to 2
+            # Mark tile as processed
             tile.seen = 2
```
choosehappy, line 119: normally you want to define these as constants to make the code a bit easier to read for 3rd parties. i.e., tile.seen = qa.constants.SEEN   (or something like that)

quickannotator/api/v1/annotation/helper.py
```diff
@@ -83,11 +83,11 @@ def count_annotations_within_bbox(table, x1, y1, x2, y2):
     result = qadb.db.session.execute(stmt).scalar()
     return result
 
-def retrieve_annotation_table(image_id: int, annotation_class_id: int, is_gt: bool) -> Table:
+def retrieve_annotation_table(session, image_id: int, annotation_class_id: int, is_gt: bool) -> Table:
```
choosehappy, line 86: i can't see everything from this scope in the code review - but is there a reason to pass in session here as a parameter versus creating it dynamically? is this function being called within the scope of some type of larger tasks which might need rollback?

PR #58, opened 2025-01-15
added tile_id as column to Annotation and hasgt as column to Tile

jacksonjacobs1:
Note that this change decreases the efficiency of adding a new annotation to the backend by ~3ms, due to the additional requirement of getting the tileId from the database. 

The added overhead for inserting 1,000,000 annotations then becomes `0.003 * 1,000,000 / 60 = 50 minutes`


quickannotator/populate_db.ipynb
```diff
@@ -214,16 +213,24 @@
...
-    "            \"isgt\": gtpred == \"gt\",\n",
-    "            \"centroid\": poly.centroid.wkt,  # Adding SRID=0 for pixel-based coordinates\n",
-    "            \"area\": poly.area,  # The area of the polygon\n",
-    "            \"polygon\": poly.wkt,  # Adding SRID=0 for the polygon\n",
-    "            \"custom_metrics\": json.dumps({\"iou\": 0.5})  # Convert custom_metrics to a JSON string\n",
-    "        }\n",
-    "        annotations.append(d)\n",
+    "        x = poly.centroid.x\n",
+    "        y = poly.centroid.y\n",
+    "        tiles = tiles_within_bbox(db, image_id, annotation_class_id, x, y, x, y)\n",
```
choosehappy, line 218: this doesn't seem optimal. we *always* know the correct tile without having to do any query. i would as well suggest removing the GEOM polygon entirely from the tile annotation table. the correct tile id is:

```python
tile_row = query_col // tile_size
tile_col = query_row // tile_size
tileid = ntile_per_col*tile_row+ tile_col
```

this should remove the database queries and all downstream needs for any tile spatial query

quickannotator/populate_db.ipynb
```diff
@@ -268,14 +275,22 @@
     "    for i, d in enumerate(tqdm(data)):\n",
     "        if d['properties']['classification']['name'] == 'tubule':\n",
     "            shapely_geometry = shape(d['geometry'])\n",
+    "            x = shapely_geometry.centroid.x\n",
+    "            y = shapely_geometry.centroid.y\n",
+    "            with app.app_context():\n",
+    "                tiles = tiles_within_bbox(db, image_id, annotation_class_id, x, y, x, y)\n",
```
choosehappy, line 281: same comment here

quickannotator/populate_db.ipynb
```diff
@@ -214,16 +213,24 @@
...
+    "        tiles = tiles_within_bbox(db, image_id, annotation_class_id, x, y, x, y)\n",
+    "        if len(tiles) == 0:\n",
+    "            annotations.append(None)\n",
+    "        else:\n",
+    "            tile_id = tiles[0].id\n",
+    "            d = {\n",
+    "                \"image_id\": image_id,\n",
+    "                \"annotation_class_id\": annotation_class_id,\n",
+    "                \"isgt\": gtpred == \"gt\",\n",
+    "                \"centroid\": poly.centroid.wkt,  # Adding SRID=0 for pixel-based coordinates\n",
```
choosehappy, line 227: do we need to store the centroid or should it be computed dynamically by the backend?

quickannotator/populate_db.ipynb
```diff
@@ -462,30 +478,19 @@
...
    "source": [
-    "insert_existing_annotations(app, db,\n",
-    "                            image_id=1,\n",
-    "                            annotation_class_id=2,\n",
-    "                            gtpred=\"gt\",\n",
-    "                            filepath='quickannotator/data/test_ndpi/13_266069_040_003 L02 PAS.json'\n",
-    "                            )"
+    "%%timeit\n",
+    "with app.app_context():\n",
+    "    print(tiles_within_bbox(db, 2, 2, 40000, 40000, 40000, 40000))"
```
choosehappy, line 493: looked at the function - not very efficient. this can also be done without any DB calls or spatial search, will be much faster like so (sketch from chatgpt, but you get the gist), based on the same idea as above:

```python
def get_tile_ids(tile_size, bbox, grid_width):
    # Bounding box: (min_x, max_x, min_y, max_y)
    min_x, max_x, min_y, max_y = bbox
    
    # Calculate the number of tiles per row
    tiles_per_row = grid_width // tile_size
    
    # Determine the tile range
    start_col = min_x // tile_size
    end_col = math.ceil(max_x / tile_size) - 1
    start_row = min_y // tile_size
    end_row = math.ceil(max_y / tile_size) - 1
    
    # Collect tile IDs
    tile_ids = []
    for row in range(start_row, end_row + 1):
        for col in range(start_col, end_col + 1):
            tile_id = row * tiles_per_row + col
            tile_ids.append(tile_id)
    
    return tile_ids

# Example usage
tile_size = 2048
bbox = (10023, 12231, 0, 5000)
grid_width = 32768  # Width of the entire grid in pixels
tile_ids = get_tile_ids(tile_size, bbox, grid_width)
print(tile_ids)
```

choosehappy (changes requested):
I think we can drop the GEOM from tile entirely, as suggested in the code, which will give you back the performance you mentioned was lost. Advent of code style ;-)

quickannotator/db/__init__.py
```diff
@@ -87,6 +87,7 @@ class Tile(db.Model):
     # columns
     geom = Column(Geometry('POLYGON'))
```
choosehappy, line 88: this GEOM can be removed, but you're missing the addition of a new column here called tile_id, which is a *relative* tile_id per image (as computed mentioned in the previous review). this field should be indexed jointly with imageid as a key (imageid + tileid). this field should also be the one stored in the annotation table and not "id", which is global. this saves a lot of querying/joining, as using the equations in the previous review, one can *always* quickly figure out what the tile_id of a particular annotation is, without having to rely on the Tiles table. do you see what i mean?

choosehappy (approved):
beautiful, looks great. no comments

PR #61, opened 2025-01-15
Mount local code to container

jacksonjacobs1:
Also changed in this PR:
1. add docker runArgs to expose all GPUs and set shm-size
2. upgrade the cuda base image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,8 +44,8 @@ RUN /opt/venv/bin/pip install tqdm \
                             ipykernel
 
 # Install node dependencies
-WORKDIR /opt/QuickAnnotator/quickannotator/client
-RUN npm install --prefix /opt/node_modules .
-ENV NODE_PATH=/opt/node_modules/node_modules
+WORKDIR /opt/
+RUN npm install QuickAnnotator/quickannotator/client
+ENV NODE_PATH=/opt/node_modules
 
 WORKDIR /opt/QuickAnnotator
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM nvidia/cuda:11.0.3-cudnn8-devel-ubuntu20.04
+FROM nvidia/cuda:12.4.1-cudnn-devel-ubuntu20.04
+
 CMD nvidia-smi
 
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" --no-install-recommends \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM nvidia/cuda:11.0.3-cudnn8-devel-ubuntu20.04
+FROM nvidia/cuda:12.4.1-cudnn-devel-ubuntu20.04
```
choosehappy, line 1: do you want 20.04? there are more recent versions which may be clever given our recent discussions:

[12.6.3-cudnn-devel-ubuntu24.04](https://hub.docker.com/layers/nvidia/cuda/12.6.3-cudnn-devel-ubuntu24.04/images/sha256-0f8250615943f311785f9ce6379a49520a4b53c124d22b42ba859edf93af3991)

or 

[12.6.3-cudnn-devel-ubuntu22.04](https://hub.docker.com/layers/nvidia/cuda/12.6.3-cudnn-devel-ubuntu22.04/images/sha256-cb239b67719dfa32ec6b525b54c1b78559bebd51a47249dd702f6c5429372154)

Dockerfile
```diff
@@ -32,19 +33,19 @@ RUN mkdir -p /opt/QuickAnnotator
...
 COPY . /opt/QuickAnnotator
 
-ENV PATH="/opt/QuickAnnotator/venv/bin:$PATH"
-
+ENV PATH="/opt/venv/bin:$PATH"
 
-RUN python3 -m venv venv \
-    && python3 -m pip install --upgrade pip \
-    && pip install -e .
+RUN python3 -m venv /opt/venv \
```
choosehappy, line 38: question - if we have a docker container which is already "isolated", what is the added value of as well using a python venv instead of installing in the "base docker o/s"? i always do the latter but am now wondering if that isn't a good practice?

choosehappy (commented):
2 comments

Dockerfile
jacksonjacobs1: It's good to have for development, and (I don't think) causes any performance degradation. 

But I must admit that it's mostly there because of previous Histotools dockerfiles.

Dockerfile
jacksonjacobs1: Sure, I don't see why not. The images are about the same size. Will commit this change directly to the v2.0 branch.

Dockerfile
choosehappy: hmm okay... i agree no performance degradation imparted by it.  i just always think that each additional piece adds complexity and opportunity for things to go astray. if you wanted to remove it, i'd be fine with it

jacksonjacobs1: Okay. I think I'm fine with it staying. The advantage of venv is that we're a little less coupled to the base docker image - i.e., if we change the base image it's possible that the default python environment changes. With a virtual environment we get a fresh environment every time.

choosehappy: smart. makes sense to me, especially given the headaches i had last night with docker : )  agree with you - lets keep it!

PR #65, opened 2025-02-04
Dynamically add tiles

jacksonjacobs1:
This PR implements the following changes:
1. No longer use spatial queries for getting tiles or annotations
2. Dynamically create tiles when ground truths and predictions are added

**Note**:
This PR does not yet account for tiles which do not intersect the tissue mask, as we can no longer utilize sqlalchemy to perform an intersection filter. This has implications for getting tile_ids to predict on, as well as for saving ground truth annotations outside of the tissue mask, but should not impact dl component development.

My plan is to work on remedies to these issues in a future PR so as not to create a bottleneck for dl component work.


quickannotator/api/v1/tile/helper.py
```diff
@@ -66,26 +76,38 @@ def get_tile_ids_within_bbox(tile_size: int, bbox: tuple, image_width: int, imag
...
-        raise ValueError(f"Point {point} is out of image dimensions (0, 0, {grid_width}, {grid_height})")
+def get_tile_id_for_point(tile_size: int, x: int, y: int, image_width: int, image_height: int) -> int:
+    if not (0 <= x < image_width and 0 <= y < image_height):
+        raise ValueError(f"Point {x}, {y} is out of image dimensions (0, 0, {image_width}, {image_height})")
 
-    x, y = point
     col = x // tile_size
     row = y // tile_size
-    tile_id = row * (grid_width // tile_size) + col
+    tile_id = row * math.ceil(image_width / tile_size) + col + 1  # Adjust tile_id to start at 1
```
choosehappy, line 93: is there or whats the motivation for it to start at 1 instead of 0? this can be a bit dangerous since we miss a +1 somewhere and its hard to track down

quickannotator/db/__init__.py
```diff
@@ -102,24 +105,24 @@ class Tile(db.Model):
...
     tile_id = Column(Integer, ForeignKey('tile.tile_id'), nullable=True, default=None)
-    
 
     # columns
     isgt = Column(Boolean, nullable=True, default=None)
-    centroid = Column(Geometry('POINT'))
+    centroid = Column(Geometry('POINT', srid=4326))  # Stored as geometry
     area = Column(Float)
-    polygon = Column(Geometry('POLYGON'))
+    polygon = Column(Geometry('POLYGON', srid=4326))  # Stored as geometry
```
choosehappy, line 122: lets keep this srid in mind - i had some serious issues previously. if you e.g., do a query without setting SRID i think it fails silently because it basically says "nope, nothing in the same space as this". i think postgis allows for leaving blank (defaults to -1 i believe) which is beautiful - i think spatialite is more particular

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: All other id indices in our db start at 1, so I figured this would be more consistent given that tileId is being used for indexing . I can change it back if you prefer. Thoughts?

quickannotator/api/v1/annotation/__init__.py
```diff
@@ -73,70 +81,80 @@ class Annotation(MethodView):
...
-        result = qadb.db.session.query(model).filter_by(id=ann.id).first()
-        return result, 200
+        
+        # Get the tile id.
+        # NOTE: The client is aware of the tilesize and image dimensions. Consider passing this information in the request or even calculating the tile_id client-side.
+        image: Image = get_image_by_id(image_id)
+        annotation_class: AnnotationClass = get_annotation_class_by_id(annotation_class_id)
+        tile_id = get_tile_id_for_point(annotation_class.tilesize, poly.centroid.x, poly.centroid.y, image.width, image.height)
+        
+        ann = insert_new_annotation(qadb.db.session, image_id, annotation_class_id, args['is_gt'], tile_id, poly)
```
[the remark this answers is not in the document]
jacksonjacobs1: Looks like flask-sqlalchemy doesn't have any rollbacks built in.

I think this would be resolved by the try-catch rollback logic to be implemented when we move to a custom SQLAlchemy session pool instead of Flask-sqlalchemy.

quickannotator/api/v1/tile/helper.py
```diff
@@ -7,56 +7,66 @@
...
+    x2 = max(0, min(bbox[2], image_width))
+    y2 = max(0, min(bbox[3], image_height))
 
     # Verify that the bounding box is within the image dimensions
-    if not (0 <= x1 < x2 <= image_width and 0 <= y1 < y2 <= image_height):
-        raise ValueError("Bounding box is out of image dimensions")
-    
+    if not (x1 < x2 and y1 < y2):
+        raise ValueError(f"Bounding box coordinates must be monotonically increasing: {bbox}")
+
```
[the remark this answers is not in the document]
jacksonjacobs1: This was a rare prospective case, but it very quickly helped me resolve an issue with the client's coordinate space where  e.g., (when zoomed out) `viewportBounds.top` was 0 and `viewportBounds.bottom` was -1 * image_height

quickannotator/api/v1/tile/helper.py
choosehappy: i think the other ids are autoincrement primary keys, so they are governed by the DB and not by us? in this case, this is a "computed" tile, which has a specific meaning.  i'm thinking about this outloud.... this ID # is actually relative to the _image_ and not to a _database table_. so if i were hacking i'd expect IDs to follow normal image conventions, upper left is (0,0) and not (1,1). contrarily, i don't think we ever explicitly use the db primary keys ids? they're mostly as results of database queries to identify unique rows, so +/- 1 is insignificant in that use case. we never e.g., iterate over database rows by index in a range(n_Row) statement, and instead iterate over result sets of rows. all in all.... i would say either to put it to 0 since that is "natural", and/or, we make a $20 bet that before the project ends, someone forgets to add a +1 somewhere where its needed and we lose an  hour debugging ;-)

quickannotator/api/v1/tile/helper.py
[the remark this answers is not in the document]
choosehappy: ha! clever!

quickannotator/api/v1/annotation/__init__.py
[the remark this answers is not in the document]
choosehappy: hmmm i think it does if you use this pattern:

```python
with Session(engine) as session:
    session.begin()
    try:
        session.add(some_object)
        session.add(some_other_object)
    except:
        session.rollback()
        raise
    else:
        session.commit()
```

jacksonjacobs1:
@choosehappy Was wondering if you had any input regarding the note at the top.

Now that we store tiles dynamically, this changes the way we get the list of tiles to predict on. We can no longer rely on a database query to compute the tiles which intersect the tissue mask. In fact, the tiles in the database are a small subset of those which actually need to be predicted on.

It seems like there are a few options 
1. while getting the list of tiles to predict on, filter all tiles using the tissue mask
2. get all tiles within the bbox and create a prediction job for each - each job will check for intersection with the mask before proceeding.

In both of these cases we'll need to check tiles for intersection, so I wrote this method for some unoptimized benchmarking:

```python
def tile_intersects_mask(image_id: int, annotatation_class_id: int, tile_id: int) -> bool:
    image = get_image_by_id(image_id)
    tilesize = get_annotation_class_by_id(annotatation_class_id).tilesize

    bbox = get_bbox_for_tile(tilesize, tile_id, image.width, image.height)
    model = create_dynamic_model(build_annotation_table_name(image_id, annotation_class_id=1, is_gt=True))
    mask_annotations = db.session.query(model).all()

    if mask_annotations:
        bbox_polygon = Polygon([(bbox[0], bbox[1]), (bbox[2], bbox[1]), (bbox[2], bbox[3]), (bbox[0], bbox[3])])
        for ann in mask_annotations:
            if bbox_polygon.intersects(wkb.loads(ann.polygon.data)):
                return True
    
    return False
```

Using %%timeit this funtion took 18ms on average to complete. I think this makes option 1 infeasible because we'll see a delay of 18ms * n_tiles before predictions can be rendered. Do you have any thoughts?

choosehappy:
hmmm....this seems rather heavy handed, and you're hitting into the some performance issues. i think there is a more elegant way since the tiles themselves are on a well established grid. 

assuming the geojson mask for the tissue is at 40x, could you imagine

1. load geojson
2. divide each point by the tile size e.g., 2048. round based on .5 (to be checked) 
3. use cv2 to draw the mask
4. use non_zero to find filled points https://numpy.org/doc/2.1/reference/generated/numpy.nonzero.html 
5. you then have r,c coordinates for the tiles, just need to do the multiplication + addition to convert into the tile_id space?

should be quite quick? would then suggest also storing this list of tiles in e.g., memcached to improve downstream speed?

quickannotator/api/v1/tile/helper.py
choosehappy: yyyyyyyyyyyyyyyyyyyyyyyup, almost *just* stepped into this in the DL component, where we have tileID and need to "inverse" compute the row / col from it. i guess that value is actually a +1, so now i need to remember everywhere to _minus_ one from it in order to get the correct r,c values. if i hadn't responded to this thread yesterday i definitely would have forgotten. lets go ahead and make it 0-based  instead of 1-based, unless you feel strongly about it?

jacksonjacobs1:
I like this approach. I'll try implementing it today.

choosehappy:
> I like this approach. I'll try implementing it today.

great! sounds good...

hmm..maybe dilate by one pixel just to make sure there are no missed edges? the additional compute overhead would be fairly low....lets see if the math works "perfectly" or only "approximately" : )

quickannotator/api/v1/annotation/__init__.py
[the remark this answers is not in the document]
choosehappy: that seems like the right thing to do - but i'm wondering if its a huge lift and/or if other folks have done something similar? i.e. ,sqlalchemy + flask aren't exactly "rarely" used projects. maybe its worth looking around a bit and seeing how others have done the same thing and/or if there are better ways and/or code for us to borrow?

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Okay, I'll push an update resolving this shortly

jacksonjacobs1:
The new method works pretty well. I used a floor operation instead of round after scaling each point. 

Comparing tile id sets between the new and old method and got the following results for our test WSI:
```
tiles_added: 0 
tiles_missed: [2225, 1650, 1859, 613]
tiles_shared: 302
```

I also put in a parameter for dilating the polygon. Only after dilating by ~250 did the new method stop missing tiles:
```
tiles_added: 20
tiles_missed: []
tiles_shared: 306
```

Kind of a weird result, no?

quickannotator/api/v1/tile/helper.py
```diff
@@ -93,9 +94,17 @@ def get_tile_id_for_point(tile_size: int, x: int, y: int, image_width: int, imag
 
     col = x // tile_size
     row = y // tile_size
-    tile_id = row * math.ceil(image_width / tile_size) + col + 1  # Adjust tile_id to start at 1
+    tile_id = row * math.ceil(image_width / tile_size) + col
```
choosehappy, line 97: hmmm...you define the get_tile_id_for_rc_index function below, why not use it everywhere, like here?

quickannotator/api/v1/tile/helper.py
```diff
@@ -93,9 +94,17 @@ def get_tile_id_for_point(tile_size: int, x: int, y: int, image_width: int, imag
...
     return tile_id
 
+def get_tile_id_for_rc_index(tile_size: int, row: int, col: int, image_width: int, image_height: int) -> int:
```
choosehappy, line 100: note that in one of my DL scripts, i put a comment which says we'll as well need the inverse function. given a tileid convert to row/col

quickannotator/api/v1/tile/helper.py
```diff
@@ -127,6 +136,39 @@ def tile_intersects_mask(image_id: int, annotatation_class_id: int, tile_id: int
...
+    # Load GeoJSON mask (assuming polygon)
+    model = create_dynamic_model(build_annotation_table_name(image_id, 1, is_gt=True))
+    mask_geojson = get_annotation_query(model).all()
+
+    polygons = []
+
+    for annotation in mask_geojson:
+        polygon = shapely.from_geojson(annotation.polygon)
+        dilated_polygon = polygon.buffer(mask_dilation)
+        scaled_polygon = np.floor((np.array(dilated_polygon.exterior.coords).astype(np.float64)) / tilesize).astype(np.int32)
```
choosehappy, line 152: probably safer to use the shapely approach since this is a shapely object

[ affinity.scale](https://shapely.readthedocs.io/en/2.0.4/manual.html#shapely.affinity.scale)

choosehappy (changes requested):
few points

choosehappy:
> The new method works pretty well. I used a floor operation instead of round after scaling each point.
> 
> Comparing tile id sets between the new and old method and got the following results for our test WSI:
> 
> ```
> tiles_added: 0 
> tiles_missed: [2225, 1650, 1859, 613]
> tiles_shared: 302
> ```
> 
> I also put in a parameter for dilating the polygon. Only after dilating by ~250 did the new method stop missing tiles:
> 
> ```
> tiles_added: 20
> tiles_missed: []
> tiles_shared: 306
> ```
> 
> Kind of a weird result, no?

i had thought about this - my intuition is that you basically want to "round down" for points on the left edge while "round up" for points on the right edge, otherwise you may end up chopping on either side. this was the reasoning for the "dilation" approach, since it addresses that a bit more organically and robustly. 

however, yes, kind of a weird result. would definitey look at a binary mask overlay to see what is/isn't included. this would be very confused for a user - feel like we need to make sure its rock solid. hmm.. just had another idea, will comment in the code review

quickannotator/api/v1/tile/helper.py
```diff
@@ -66,26 +80,95 @@ def get_tile_ids_within_bbox(tile_size: int, bbox: tuple, image_width: int, imag
...
+    
+    # Load GeoJSON mask (assuming polygon)
+    model = create_dynamic_model(build_annotation_table_name(image_id, 1, is_gt=True))
+    mask_geojson = get_annotation_query(model).all()
+
+    polygons = []
+
+    for annotation in mask_geojson:
+        polygon = shapely.from_geojson(annotation.polygon)
+        dilated_polygon = polygon.buffer(mask_dilation)
```
choosehappy, line 151: actually, i would probably do this differently upon second thought, which may address the issue you were seeing in your github comment. i would

1. first downsize the polygons
2. make the binary mask
3. then use cv2 to perform the dilation

this has a few benifits:
a) i think its more robust
b) it'll be faster 1 single dilation call instead of potentially e.g., 1000 of them
c) its easier to work/visualize inthe pixel space as opposed to the coordinate space.

thoughts?

choosehappy (changes requested):
dilation approach considerations

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Yes this sounds reasonable - we'll probably end up with a lot more background tiles, but this is a fairly benign side effect - and much more desirable behavior for the user than being blocked from saving annotations somewhere.

Will make the changes now.

jacksonjacobs1: With no dilation
image

With 1 pixel of mask dilation
image

jacksonjacobs1: I think the problem stems from the improbable case where we see adjacent points in the polygon somehow have a manhattan distance of 2 - they should all have 1, right?

jacksonjacobs1: What we're seeing is this case, where by chance the points within the polygon "skip" the manhattan distance 1 tile over to the manhattan distance 2 tile.
image

I think we just need to interpolate more points, but the dilation operation also seems appropriate to me.

choosehappy: "improbable" is probably going to be pretty probable ;-) it doesn't bode well when the _first_ image we try this with has this issue. suggests it'll be a pretty common occurrence. i would say, even though the dilation is a bit heavy handed, lets do that route because its fast/easy/robust. we can always revert back and add this as a feature request for v2.1 to improve efficiency. i suspect as well there is probably a different interpolation approach that could resolve this, but would take more effort to sketch out. e.g., instead of dividing by tile_size, divide by tile_size/4, which gives more of a fine grained approach. figure out which points are positive, and then map those to the tile_size space. you'll still have the chance of missing one or two, but it'll become less and less likely as /4 -- > inf

jacksonjacobs1: Confirmed that interpolating 10x the number of points in the polygon minimizes the chance of the above situation occurring, resulting in manhattan distance = 1 for every adjacent point:
image

I first tried interpolation using shapely but this was very slow because shapely only has a method for interpolating 1 point at a time, and I stuck it in a for loop. (took almost 1 second to run... yikes)

I decided not to try optimizing using numpy function vectorization and instead used the interpolation method already written for bioptort. this added the requirement of scipy, but got the run time down to 8ms.

jacksonjacobs1: Just saw your comment.

Actually I think interpolation is more robust, because in qupath a user can annotate a polygon by "clicking", resulting in a mask with very few points on the polygon. This mask could skip tiles completely. Of course this is a rare use case, but if we interpolate points as a function of tilesize we should be able to avoid this issue entirely.

jacksonjacobs1: Thought about it some more - even in the above case, cv2's fillPoly should still work fine. I'm good to move forward without point interpolation!

jacksonjacobs1: Okay - I think I have the most robust solution -

```python
cv2.fillPoly(mask, polygons, 255, lineType=cv2.LINE_4)
```

This draws the line with 4-connectivity, preventing any diagonally-adjacent pixels.

chatgpt:

How Does `cv2.LINE_4` Decide the Next Pixel?

In **4-connected line drawing (`cv2.LINE_4`)**, the algorithm allows movement **only in horizontal and vertical directions** (not diagonally). It determines the next pixel by minimizing the error between the ideal mathematical line and the selected pixels.

jacksonjacobs1: Just pushed the final version - note that `LINE_4` only actually fixed the problem 50% of the time, since our polygon's "diagonal" sections perfectly intersect mask pixels.

I ended up keeping LINE_4 just because it's slightly more efficient than the LINE_8 default.

I then use a dilation operation with kernel:
```
010
111
010
```
...because this is a little less heavy handed than a kernel with all 1's.

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Looks fine, although numpy allows us to control the precision -  I don't think using numpy here increases the risk any more than performing a floor operation with numpy later?

jacksonjacobs1 (commented):
Everything looks good on my side to merge in.

quickannotator/api/v1/tile/helper.py
choosehappy: looks good. seems like a very reasonable first (and perhaps final) pass - seems like a trivial thing to come back and adjust later if needed. don't think we need to do any further testing/thinking about this unless we start noticing some issues, but it feels like its pretty well handled

quickannotator/api/v1/tile/helper.py
choosehappy: generally speaking, affinity.scale is somehow "safer", because its a package designed for manipulating polygons in polygon/coordinate space. it could have internally other "stuff" to make it more robust or more performant, e.g., when it downscales perhaps it reduces the # of points needed, or when it upscales, perhaps it adds. since there is a specalized library for it, its safer to use it as opposed to try and implement the same functionliaty yourself. e.g., they thought about all edge cases and many folks have used the code and debugged it and ensured that it works as expected. your code is brand new and hasn't had that luxury. so my general rule is - never write an equation yourself if an existing popular toolbox has something which implements the same thing - you'll end up saving yourself a few hours of debugging over the course of a project ;-)

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -0,0 +1,72 @@
...
+from sqlalchemy.orm import aliased, sessionmaker, Session, Query, DeclarativeBase
+import quickannotator.db as qadb
+from quickannotator.db import build_annotation_table_name, create_dynamic_model
+import shapely
+import json
+from sqlalchemy import func, text
+
+
+# TODO: Remove session from params
+def get_tile(session: Session, annotation_class_id: int, image_id: int, tile_id: int) -> qadb.Tile:
```
choosehappy, line 11: inline with previous comments, its probably session, image id, class id, tileid in order of rate of change

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -0,0 +1,72 @@
...
+        tile_id=tile_id
+    ).first()
+    return result
+
+
+def compute_custom_metrics() -> dict:
+    return {"iou": 0.5}
+
+# TODO: Remove session from params
+def insert_new_annotation(session, image_id, annotation_class_id, is_gt, tile_id, polygon: shapely.geometry.Polygon):
```
choosehappy, line 24: note how here you have a different odering of parameters as compared to above, without any really strong justification for doing so. this is mainly because as humans if we don't have strict rules, we will act fairly arbitrarily. this is easy for us, but you can imaigne if you're trying to work on this project having different orderings can be confusing and lead to unintended errors - thus its better to make a rule and stick with it, so that you don't actually _have_ to think about what to do, and instead just do the right thing directly the first time. see what i mean?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -0,0 +1,72 @@
...
+            - polygon: The annotation polygon in GeoJSON format.
+            - area: The area of the annotation polygon.
+            - custom_metrics: Custom metrics associated with the annotation.
+            - datetime: The datetime when the annotation was created or last modified.
+    '''
+
+    query = qadb.db.session.query(
+        model.id,
+        model.tile_id,
+        func.ST_AsGeoJSON(model.centroid).label('centroid'),
```
choosehappy, line 65: note that i think ST_ is for spatiliate only and not postgis - so this is DB specific code. this begs the question, how often do we use DB specific functions? if there are only 1 or 2, it likely makes sense to abstract and wrap them (if sqlite , else postgis..) so that later on we can swap out trivially without having to do a code hunt? i think you're seeing the project is growing in size - making code hunting more expensive and unforunately more likely to be error prone

quickannotator/api/v1/tile/__init__.py
```diff
@@ -35,11 +41,19 @@ class PostTileArgsSchema(Schema):
 class SearchTileArgsSchema(Schema):
     image_id = fields.Int(required=True)
     annotation_class_id = fields.Int(required=True)
+    hasgt = fields.Bool(required=False)
+    include_placeholder_tiles = fields.Bool(required=False, default=False)    # A placeholder tile is a placeholder tile that has not yet been created in the database.
```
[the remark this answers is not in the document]
jacksonjacobs1: In the current implementation we return placeholder tiles so that tiles are only added to the db when either
1. new gts have been committed 
2. /predict has been called for a tile

This change would instead mean that we create tiles within the bounding box whenever we search for tiles by bounding box.

actually I see no downsides with this approach. I've modified the code accordingly.

jacksonjacobs1:
Thanks, all valid feedback.

Unfortunately it turns out I can't commit additional code to a closed PR. I'll generate a new PR implementing the changes discussed here.

quickannotator/api/v1/tile/__init__.py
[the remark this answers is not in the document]
choosehappy: i leave it up to you - i don't think i have a full scope on the functionality, don't acquiesce : ) e.g., is a (likely) case where the bounding box takes up the entire view port (just logging into an image), and thus all tiles are automatically (almost always)?  maybe we take a step back - can i ask more broadly, in what situation does a placeholder need to exist at all? from a purely naive standpoint, if its "empty" its not containing any information, so what actually needs to consume an "empty" tile?

PR #68, opened 2025-02-07
Improve devops

### LINKED ISSUE #64: Improve devops -
## Description
Re-running the dev container currently requires an image rebuild when code changes. Additionally, pip is slow.

## Implementation notes
1. Extract dependencies from pyproject.toml into requirements.txt file. The txt file will be referenced by the toml file.
https://stackoverflow.com/questions/73600082/how-to-reference-a-requirements-txt-in-the-pyproject-toml-of-a-setuptools-projec
2. Only copy requirements.txt and node dependencies into the container.
3. Remove  `pip install -e quickannotator` from the Dockerfile and instead tell devcontainers to run this after building the image.
4. Add RUN commands to dockerfile to install and use UV instead of pip and virtualenv. 

## Acceptance criteria

- [x] Code editing no longer invalidates the cached docker image layer
- [x] uv is used for all python package management

jacksonjacobs1:
closes #64

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get install -y nodejs
 
 RUN mkdir -p /opt/QuickAnnotator
 WORKDIR /opt/QuickAnnotator
-COPY ./pyproject.toml /opt/QuickAnnotator
+COPY ./pyproject.toml /opt/QuickAnnotator/pyproject.toml
 
 # Install uv
 RUN pip install uv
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -31,21 +37,21 @@ RUN apt-get install -y nodejs
...
+RUN uv venv /opt/uv_venv
+ENV PATH="/opt/uv_venv/bin:$PATH"
+RUN uv pip install -r <(uv pip compile pyproject.toml)
 
 # Install node dependencies
+COPY ./quickannotator/client/package.json /opt/QuickAnnotator/quickannotator/client/package.json
+COPY ./quickannotator/client/package-lock.json /opt/QuickAnnotator/quickannotator/client/package-lock.json
 WORKDIR /opt/
-RUN npm install QuickAnnotator/quickannotator/client
+RUN npm install QuickAnnotator/quickannotator/client/
```
choosehappy, line 54: does "npm ci" save us some speed?

jacksonjacobs1: It does give us a speed boost - oddly npm ci is different from npm install in that it does not allow for installing node_modules in a directory other than the one containing package.json and package-lock.json.

I think this is still okay, I'll just have to copy these json files to /opt before installing.

choosehappy: nnnnnnice. whats the approx timing difference?

jacksonjacobs1: Down to ~7 seconds, although I'm suspicious there's caching involved. 

I'm not sure how cacheing would be possible though because I deleted the image which should prevent the use of cached layers.

jacksonjacobs1:
This PR is ready to be merged in, unless you have any other thoughts

choosehappy:
None noted, let'er rip!

Issue #42, opened 2024-11-14
Project page - layout implementation


jacksonjacobs1:
# Project page - layout implementation

### Description 
Using react, implement the following user interface layout:

![image](https://github.com/user-attachments/assets/ad475b8f-f737-4cb4-8bc0-6a53bdc41f55)

### Feature details
- The page should be populated with the following React components:
  - imageList
  - uploadBox
  - actionsToolbar
  - confirmationPopover

### Implementation Notes
- The image list should be implemented using slickgrid to allow for sorting and infinite scrolling. Slickgrid has already been implemented in the `annotationList.tsx` component.

jacksonjacobs1:
@nanli-emory can you link to a PR resolving this feature request if it's done?

PR #69, opened 2025-02-07
Project page

nanli-emory:
implement the following user interface layout which includes imageList, uploadBox, actionsToolbar, and confirmationPopover.(https://github.com/choosehappy/QuickAnnotator/issues/42)


quickannotator/api/v1/image/__init__.py
```diff
@@ -91,19 +92,67 @@ def get(self, args, project_id):
...
+        """
+        # images = db.session.query(qadb.Image).filter(qadb.Image.project_id == project_id).all()
+        # if images is not None:
+        #     return images
+        # else:
+        #     abort(404, message="Images not found")
+        images = db.session.query(
+            *[getattr(qadb.Image, column.name) for column in qadb.Image.__table__.columns],
+            db.func.ST_AsGeoJSON(qadb.Image.embedding_coord).label('embedding_coord')
+        ).filter(qadb.Image.project_id == project_id).all()
```
choosehappy, line 111: if its easier, we can think about storing these as simple a tuple instead of a GEOM

quickannotator/api/v1/image/__init__.py
```diff
@@ -91,19 +92,67 @@ def get(self, args, project_id):
...
 #################################################################################
+@bp.route('/upload/files', methods=["POST"])
+def upload_files():
+    """Upload a file"""
+    filenames=[]
+    files = request.files.getlist('files')
+    
+    for file in files:
+        filename = secure_filename(file.filename)
+        os.makedirs('./upload', exist_ok=True)
```
choosehappy, line 126: need to thikn about the directory sturcutre of the project and place this correctly. should be within the same project directory and then use just a "rename" to put it in place

quickannotator/api/v1/image/__init__.py
```diff
@@ -91,19 +92,67 @@ def get(self, args, project_id):
...
+        full_path = os.path.join(current_app.root_path, result.path)
         if file_type == 1:  # image file
-            return send_from_directory(result['path'], result['name'])
+            return send_from_directory(full_path, result.name)
         elif file_type == 2:    # thumbnail file
-            # TODO implement thumbnail file
-            pass
+            try:
+                slide = ops.OpenSlide(full_path)
+                thumbnail = slide.get_thumbnail((256, 256))
```
choosehappy, line 146: we should think about the time cost here - if these images are on e.g., S3, and there are 1,000 images, is this going to take a long time? also, a 256 x 256 thumbnail is very small, so we could easily consider saving it in the project directory so that the first time its slow but all other times its very fast since it will be loaded directly from disk?

choosehappy (changes requested):
some comments to consider in the python part - maybe @jacksonjacobs1 can review the JS ?

.github/workflows/build_v2_docker_image.yml
```diff
@@ -0,0 +1,46 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+# GitHub recommends pinning actions to a commit SHA.
+# To get a newer version, you will need to update the SHA.
+# You can also reference a tag or branch, but the action may change without warning.
+
+name: Publish Docker image
```
jacksonjacobs1, line 10: Please see #68 which breaks this docker image build. Our dockerfile no longer actually copies the full quickannotator package into the container. Instead, the code is mounted into the container from the local directory, as specified in the `devcontainer.json` file.

If you want to build a CI for quickannotator, I would recommend writing a second dockerfile which copies in the code. However, we might want to hold off altogether since we may end up using a docker compose file instead. Thoughts?

PR #70, opened 2025-02-12
Decouple Flask-sqlalchemy from application.

jacksonjacobs1:
This PR removes Flask-SQLAlchemy as a requirement and instead uses SQLAlchemy's `scoped_session` for thread-local session management.

This is in-line with the "Declarative" approach discussed here:
https://flask.palletsprojects.com/en/stable/patterns/sqlalchemy/

We also use contextmanager (with a custom `get_session()` function) to safely commit, rollback, and remove sessions outside of flask application request threads.

The pattern for getting sessions within a flask request is as follows:
```python
from quickannotator.db import db_session

@bp.route('/<int:image_id>/<int:annotation_class_id>')
class Annotation(MethodView):
    def get(self, args, image_id, annotation_class_id):
        return db_session.query(model).all()
```

And the pattern for getting sessions outside of the flask application is:

```python
from quickannotator.db import get_session

with get_session() as db_session:
    db_session.query(model).all()

```

You can also use utility functions defined within the flask application, as long as you use the get_session context manager to clean up the session:

```python
from quickannotator.db import get_session
from quickannotator.api.v1.utils.shared_crud import get_tile

with get_session() as db_session:
    tile = get_tile(2, 1, 1)

### CODE REVIEW

File: quickannotator/api/v1/image/__init__.py [hunk 1]
```diff
@@ -53,10 +51,10 @@ class Image(MethodView):
     def get(self, args):
         """     returns an Image
         """
-        result = db.session.query(
-            *[getattr(qadb.Image, column.name) for column in qadb.Image.__table__.columns],
-            db.func.ST_AsGeoJSON(qadb.Image.embedding_coord).label('embedding_coord')
-        ).filter(qadb.Image.id == args['image_id']).first()
+        result = db_session.query(
+            *[getattr(quickannotator.db.models.Image, column.name) for column in quickannotator.db.models.Image.__table__.columns],
+            func.ST_AsGeoJSON(quickannotator.db.models.Image.embedding_coord).label('embedding_coord')
```
choosehappy, line 56: similar to my other comment in previous review, ST_ i believe is sqlite only and not postgis? maybe abstrac tion is warranted? (or i might be wrong, do both support it?). another potentially more interesting idea - submit a pull request to their repos which deal with this internally? e.g., modify geoalchemy to do this abstraction on our behalf, this way we don't have to?

quickannotator/api/v1/image/utils.py
```diff
@@ -0,0 +1,26 @@
+from quickannotator.db import db_session
+import quickannotator.db.models as models
+import large_image
```
choosehappy, line 3: where'd large image come into this?

quickannotator/api/v1/project/__init__.py
```diff
@@ -56,8 +56,7 @@ def post(self, args):
         """     create a new Project
         """
 
-        db.session.add(qadb.Project(name=args['name'], description=args['description'], is_dataset_large=args['is_dataset_large']))
-        db.session.commit()
+        db_session.add(models.Project(name=args['name'], description=args['description'], is_dataset_large=args['is_dataset_large']))
```
choosehappy, line 59: is this large boolean coming from the UI where the user specifies project size? might be better to rename this from "large" to an explicit value that was used in the UI so that the code ism ore transparent? e.g., "is_dataset_larger50k" or something?

quickannotator/api/v1/tile/helper.py
```diff
@@ -191,65 +192,40 @@ def generate_random_circle_within_bbox(bbox: Polygon, radius: float) -> shapely.
...
+        bbox = get_bbox_for_tile(annotation_class.tilesize, tile_id, image.width, image.height)
+        bbox_polygon = Polygon([(bbox[0], bbox[1]), (bbox[2], bbox[1]), (bbox[2], bbox[3]), (bbox[0], bbox[3])])
+        for _ in range(random.randint(20, 40)):
+            polygon = generate_random_circle_within_bbox(bbox_polygon, 100)
+            insert_new_annotation(image_id, annotation_class_id, is_gt=False, tile_id=tile_id, polygon=polygon)
 
-    except Exception as e:
-        print(f"Error processing tile: {e}")
+        # Mark tile as processed
+        tile.seen = 2
```
choosehappy, line 218: should use ENUM

quickannotator/api/v1/tile/__init__.py
```diff
@@ -109,27 +109,27 @@ class TileSearch(MethodView):
...
-            query = query.filter(qadb.Tile.hasgt == args['hasgt'])
+            query = query.filter(models.Tile.hasgt == args['hasgt'])
         
         tiles = query.all()
 
         if args['include_placeholder_tiles']:
             existing_ids = set([tile.tile_id for tile in tiles])
             placeholder_tile_ids = ids - existing_ids
-            placeholder_tiles = [qadb.Tile(tile_id=tile_id, image_id=args['image_id'], annotation_class_id=args['annotation_class_id'], seen=0, hasgt=False) for tile_id in placeholder_tile_ids]
+            placeholder_tiles = [models.Tile(tile_id=tile_id, image_id=args['image_id'], annotation_class_id=args['annotation_class_id'], seen=0, hasgt=False) for tile_id in placeholder_tile_ids]
```
choosehappy, line 132: enum for seen

quickannotator/db/models.py
```diff
@@ -0,0 +1,176 @@
...
+        kwargs["metadata"] = {
+            "type": "string",  # You can also specify "object" if it's GeoJSON
+            "description": "A geometry field. Serialized WKB into geojson."
+        }
+        super().__init__(*args, **kwargs)
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        if value is None:
+            return None
+        # if isinstance(value, str):
```
choosehappy, line 167: does this need to be defined?

choosehappy (changes requested):
looks great, nice job! very clean. a few minor comments

quickannotator/api/v1/image/utils.py
jacksonjacobs1: I think we previously discussed this - I chose large image for serving WSI tiles / reading regions because
1. it wraps openslide while also adding guardrails logic for slides with missing levels and metadata - these were problems I ran into while trying out the openslide tileserver example.
2. it is modular, e.g., we can pip install large_image[bioformats] etc., to easily add support for formats which openslide does not support
3. package overhead was relatively low - I forget how much but I think it was less than 20mb
4. It was the tile-serving library of choice for geojs and DSA and I was able to quickly get the tile server up and running in just a few lines of code.

quickannotator/api/v1/project/__init__.py
jacksonjacobs1: This was kind of a loose placeholder because we didn't brainstorm functionality that would fundamentally change based on the project size. 

I think we discussed warning the user that large projects may require more compute, i.e., a larger ray cluster?

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Are you referring to using the ENUM sqlalchemy type? Or more shallowly to use python's enum to refer to these values in the backend?

choosehappy: the latter, something more like a "constants.py" which defines all numerical constants globally throughout the project. do you see what i mean?

quickannotator/db/models.py
jacksonjacobs1: Nope! Took it out.

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Yes, done. Can you take a look?

quickannotator/api/v1/image/__init__.py
jacksonjacobs1: Hmm - ST_AsGeoJSON seems to be working for Spatialite as is - the database response is in geojson, despite using a spatialite db.

It's unclear in the Geoalchemy documentation, but I think ST_AsGeoJSON maps to AsGeoJSON in spatialite:
https://github.com/geoalchemy/geoalchemy2/issues/202

jacksonjacobs1: Also see
https://github.com/geoalchemy/geoalchemy2/pull/204

quickannotator/api/v1/project/__init__.py
choosehappy: yes, thats right, thanks for the reminder. generally i just take issue with the vagueness of "large", since your definition of large may not be the same for that of the user. this definitely means that we'll need to include clear instructions in the front end (obviously, and i believe thats noted in the figma as well), but i worry that e.g., a backend developer won't know what this means since they won't read the front end code. for stuff like that, which is ambigious in nature, its often customary to add a comment where the variable is defined so folks can "stumble" upon the relevant information without having to wonder. so maybe we just add a note in the model definition and call it a day?

quickannotator/api/v1/image/__init__.py
choosehappy: got it, yes, you're right, just tested this out. i remember much more clearly now from my previous tests. 

ST_AsGeoJSON doesn't work/exist when using Core, but does exist if you use it as a function and/or ORM:


```python
import sqlite3

# Connect to an in-memory SpatiaLite database
conn = sqlite3.connect(":memory:")
conn.enable_load_extension(True)
conn.execute("SELECT load_extension('mod_spatialite')")

# Create a spatial table and insert a geometry
conn.execute("CREATE TABLE spatial_test (id INTEGER PRIMARY KEY, geom GEOMETRY)")
conn.execute("INSERT INTO spatial_test (geom) VALUES (GeomFromText('POINT(10 20)', 4326))")

# Test ST_AsGeoJSON
cursor = conn.execute("SELECT ST_AsGeoJSON(geom) FROM spatial_test")
print("SpatiaLite:", cursor.fetchone()[0])

conn.close()
```

which yields this error:

```python
---------------------------------------------------------------------------
OperationalError                          Traceback (most recent call last)
Cell In[3], line 13
     10 conn.execute("INSERT INTO spatial_test (geom) VALUES (GeomFromText('POINT(10 20)', 4326))")
     12 # Test ST_AsGeoJSON
---> 13 cursor = conn.execute("SELECT ST_AsGeoJSON(geom) FROM spatial_test")
     14 print("SpatiaLite:", cursor.fetchone()[0])
     16 conn.close()

OperationalError: no such function: ST_AsGeoJSON
```

however, this "func" approach with geoalchemy works as expected:

```python
from sqlalchemy import create_engine, event, Column, Integer, text
from sqlalchemy.orm import declarative_base, sessionmaker
from geoalchemy2 import Geometry
from sqlalchemy.sql import func

# Create an SQLite engine
engine = create_engine("sqlite:///test.db", echo=True)

# Event listener to load SpatiaLite extension on each new connection
@event.listens_for(engine, "connect")
def load_spatialite(dbapi_connection, connection_record):
    dbapi_connection.enable_load_extension(True)
    try:
        # Load the SpatiaLite extension; adjust the path if necessary.
        dbapi_connection.load_extension("mod_spatialite")
    except Exception as e:
        print("Error loading SpatiaLite extension:", e)

# Initialize the spatial metadata
with engine.connect() as conn:
    conn.execute(text("SELECT InitSpatialMetadata(1);"))

Base = declarative_base()

# Define the ORM model
class SpatialTest(Base):
    __tablename__ = "spatial_test"
    id = Column(Integer, primary_key=True)
    geom = Column(Geometry("POINT", srid=4326))

# Create the table
Base.metadata.create_all(engine)

# Insert a test point
Session = sessionmaker(bind=engine)
session = Session()
session.execute(
    SpatialTest.__table__.insert().values(geom="SRID=4326;POINT(10 20)")
)
session.commit()

# Query using ST_AsGeoJSON, using .first() to avoid MultipleResultsFound error
result_row = session.query(func.ST_AsGeoJSON(SpatialTest.geom)).first()
geojson_result = result_row[0] if result_row is not None else None
print("\t\tSpatiaLite (GeoAlchemy2):", geojson_result)

session.close()

```

yielding

```python
SpatiaLite (GeoAlchemy2): {"type":"Point","coordinates":[10,20]}
```

so i guess if we're planning on ever writing any native sql code, we need to be very very careful about the underlying functions - however if we use the "alchemies", we're ok

jacksonjacobs1: Thanks for confirming. Yes if all goes well we can avoid using core altogether :)

quickannotator/api/v1/tile/helper.py
choosehappy: _very_ close, but i think we'd be better off using an IntEnum

https://docs.python.org/3/library/enum.html#enum.IntEnum

which would allow us to use:

`TileStatus.UNSEEN`

instead of

`TileStatus.UNSEEN.value`



worth taking a look and familiarizing yourself with the other types:
https://docs.python.org/3/howto/enum.html#derived-enumerations

interestingly - i didn't know these existed 10 minutes ago, but the ".value" set me off on a path of discovery :)

quickannotator/api/v1/image/utils.py
jacksonjacobs1: I guess it follows that we should use large_image throughout the system - I can see switching from openslide to large_image is a little inconvenient given that the interface is a little different (`getRegion()` has different params vs `read_region()`).

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Just took a look, nice find. I've gone and updated the code

jacksonjacobs1: Note that the merge commit was necessary due to some conflicts in imports from the dl component. Will need to update the dl component once this PR is merged.

jacksonjacobs1:
I've also committed code implementing changes requested in #65 after that PR was already merged.

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Great, thanks! This is now done.

PR #74, opened 2025-02-20
Add pytests

### LINKED ISSUE #71: Tests
### Description
Currently, QA has no unit tests, which makes it difficult to ensure code reliability and prevent regressions. This issue aims to add unit tests for key functionalities using `pytest`

### Tasks
- [x] Set up `pytest` if not already installed.
- [x] Add test configurations in `pytest.ini` or `conftest.py`.
- [ ] Write unit tests for the following components:
  - [x] Routes
  - [ ] Models
  - [ ] Services/Business Logic
  - [ ] Error Handling
- [ ] Ensure tests cover at least 80% of the codebase.
- [ ] Integrate test execution into CI/CD pipeline (if applicable).

### Suggested Tools
- `pytest`
- `coverage.py`

### Example tests folder structure from resource #1
```
└── tests
    ├── conftest.py
    ├── functional
    │   ├── __init__.py
    │   ├── test_books.py
    │   └── test_users.py
    └── unit
        ├── __init__.py
        └── test_models.py
```

### References
https://testdriven.io/blog/flask-pytest/
https://www.digitalocean.com/community/tutorials/unit-test-in-flask#why-pytest-is-a-better-alternative-to-unittest
https://flask.palletsprojects.com/en/stable/testing/

jacksonjacobs1:
Closes #71 

Currently only tile API routes are tested.

Note that functional tests inflate coverage values because flask route code is counted.

Overall coverage of python code is reported as 72%


quickannotator/__main__.py
```diff
@@ -1,16 +1,17 @@
+import os
+
+os.environ['TESTING'] = 'false'  # need to set this before importing the db module
```
choosehappy, line 3: hmm seems strange - this should be "False" by default if it isn't defined in the o/s? i would have expected only to require setting Testing when True - which would be done by the pytest directly?

choosehappy: also seems strange that "false" is lower case?

quickannotator/config/__init__.py
```diff
@@ -1,12 +1,16 @@
...
+import os
 
 # initialize a new config file:
 config = configparser.ConfigParser(interpolation=configparser.ExtendedInterpolation())
 
 config.read("quickannotator/config/config.ini")
 
 def get_database_uri():
-  return config.get('sqlalchemy', 'database_uri', fallback='sqlite:///quickannotator.db')
+  if os.getenv('TESTING', 'false').lower() == 'true':
```
choosehappy, line 10: a few thoughts, 1, i typically like to avoid doing this sort of comparison and instead would suggest using this:
https://docs.python.org/2/distutils/apiref.html#distutils.util.strtobool

which i think you can understand why is much nicer : )  

however, it looks like it will be deprecated - and i can't seem to find a reasonable replacement - do you see one? would be nice to know. t here is some code, but i'd prefer an estalibshed large library if possible (better support and maitenance: https://danielms.site/zet/2023/pythons-distutil-strtobool-replacement/)




this seems like a strange way to write this because your'e basicalyl saying "if false is true, then true" which hurts my brain. wouldn't a better way be, noting that it defaults to false if not set in the o/s (Addressing my comment above)

```python
testing = bool(strtobool(os.getenv("TESTING", "false")))

if testing:
    return config.get("sqlalchemy", "test_database_uri", fallback="sqlite:///:memory:")
else:
    return config.get("sqlalchemy", "database_uri", fallback="sqlite:////opt/QuickAnnotator/quickannotator/instance/quickannotator.db")
```

quickannotator/tests/conftest.py
```diff
@@ -0,0 +1,97 @@
...
+                        project_id=1,
+                        name="Fake Class",
+                        color="red",
+                        magnification=10,
+                        patchsize=256,
+                        tilesize=2048,
+                        dl_model_objectref=None)
+
+    # Add a tile
+    upsert_tile(
```
choosehappy, line 89: i would say to make sure that both the insert and the update work?

choosehappy (changes requested):
good start!

quickannotator/__main__.py
jacksonjacobs1: Looks like pytest sets the `PYTEST_CURRENT_TEST` environment variable automatically:
```python
import psutil

for pid in psutil.pids():
    environ = psutil.Process(pid).environ()
    if "PYTEST_CURRENT_TEST" in environ:
        print(f'pytest process {pid} running: {environ["PYTEST_CURRENT_TEST"]}')
```

It's probably better practice for me to use this variable directly than to manage our own variable. It also allows us to check for existence rather than casting the string to a boolean. Thoughts?

quickannotator/tests/conftest.py
jacksonjacobs1: Will do. I mostly focused on functional tests in this commit but will add some unit tests today. I don't intend to spend that much more time on test infrastructure before the MVP release - lots of other features to work on!

quickannotator/__main__.py
choosehappy: yessssssssss, love it

quickannotator/tests/conftest.py
choosehappy: yes, likely the right thing to do . however, i would be interested in seeing how well e.g., chatgpt works in a more professional development workflow. e.g., you do some coding and add functionality -- what if you ask chatgpt to immediately write a unit test to cover it?  this has two benifits (a) no context switching penalty, (b) i suspect you'll already find/fix bugs in the new functionality which may have been hidden overall saving time, (c) overall in the end, with maybe 5% additional effort, getting much better code coverage.

note that this hinges on chatgpt/co-pilot being able to make tests efficiently - this is something i think you'll need to test and report back on  :)

quickannotator/__main__.py
jacksonjacobs1: https://stackoverflow.com/questions/25188119/test-if-code-is-executed-from-within-a-py-test-session

Looks like `PYTEST_CURRENT_TEST` does not help us in our use case, because it is only set when a test is being run, not when fixtures are being set up.

I found this after confirming this behavior for myself.

This suggested solution worked for me:
```python
import sys

if "pytest" in sys.modules:
    ...
```

However we could also just set our own `TESTING` variable and check for existence. What do you prefer?

PR #78, opened 2025-02-24
Add templates for github issues and feature requests

jacksonjacobs1:


jacksonjacobs1:
also snuck in matplotlib as a dependency

PR #81, opened 2025-02-26
Work mag

### LINKED ISSUE #80: [FEATURE] Store annotations and tilesize in working mag space
## Description  
Currently annotations are stored in basemag space. Additionally, tilesize is not tightly coupled to the working magnification for a particular class.

We define a few naming conventions here:
- `work_` refers to the coordinate space of the image at the working magnification
- `base_` refers to the coordinate space of the image at the base magnification
- These are used to prefix other 

The following changes are then requested:

1. Refactor model definitions
    **AnnotationClass**
    `magnification`  --> `work_mag`
    `tilesize` --> `work_tilesize`
    
    **Image**
    `width` -> `base_width`
    `height` -> `base_height`
    
    **Annotation**
    `centroid` -> `work_centroid`
    `polygon` -> `work_polygon`

2. Refactor methods for getting tileids etc
3. Automatically scale annotations from the database using `func.ST_Scale`
4. Scale new annotations to work_mag before saving them to the db.

jacksonjacobs1:
Closes #80 

Notes:
- The column names for polygon and centroid were not changed, since the coordinate space of these fields is stored in workmag but returned in basemag
- Functions relating to tilespace conversions were added as methods to a new TileSpace class. These methods are agnostic of magnification - all inputs in the workmag space map to outputs in the workmag space.
- All Annotation geoms in the db are assumed to be stored in the `workmag` coordinate space. POST and PUT methods automatically scale Annotations into the `workmag` space.
- All polygon and centroid queries use the `ST_Scale` method to scale geometries up to the `basemag` space before  returning to the frontend.

The following db colums were renamed as discussed in #80 
**AnnotationClass**
`magnification`  --> `work_mag`
`tilesize` --> `work_tilesize`

**Image**
`width` -> `base_width`
`height` -> `base_height`

Some refactorings are still not complete - TileSpace has not yet been integrated into all methods. I wanted to get some initial feedback before moving ahead.


jacksonjacobs1:
Another thought - while TileSpace is agnostic to magnification, a wrapper method can be used for greater safety:
```python
def get_tilespace(image_id: int, annotation_class_id: int, in_work_mag: bool=True) -> TileSpace:
    image = get_image_by_id(image_id)
    annotation_class = get_annotation_class_by_id(annotation_class_id)
    r = base_to_work_conversion_factor(image_id, annotation_class_id)
    if in_work_mag:
        image_work_width = image.base_width * r
        image_work_height = image.base_height * r
        return TileSpace(annotation_class.work_tilesize, image_work_width, image_work_height)
    else:   # return
        base_tilesize = annotation_class.work_tilesize / r
        return TileSpace(base_tilesize, image.base_width, image.base_height)
```

jacksonjacobs1:
I've added a `get_tilespace` method similar to the one described above. In our backend endpoints, this method is called like so:

```python
tilespace = get_tilespace(args['annotation_class_id'], args['image_id'], in_work_mag=False)
bbox = tilespace.get_bbox_for_tile(args['tile_id'])
```
... where the bbox is in the base_mag space

In the DL code, you can instead call

```python
tilespace = get_tilespace(args['annotation_class_id'], args['image_id']) # in_work_mag is True by default
bbox = tilespace.get_bbox_for_tile(args['tile_id'])
```

... where the bbox is returned in the work_mag space.

choosehappy:
> I've added a `get_tilespace` method similar to the one described above. In our backend endpoints, this method is called like so:
> 
> ```python
> tilespace = get_tilespace(args['annotation_class_id'], args['image_id'], in_work_mag=False)
> bbox = tilespace.get_bbox_for_tile(args['tile_id'])
> ```
> 
> ... where the bbox is in the base_mag space
> 
> In the DL code, you can instead call
> 
> ```python
> tilespace = get_tilespace(args['annotation_class_id'], args['image_id']) # in_work_mag is True by default
> bbox = tilespace.get_bbox_for_tile(args['tile_id'])
> ```
> 
> ... where the bbox is returned in the work_mag space.

hmmmm the question is -- is this actually needed? i think we're storing the tile as a GEOM in the tile table, if we store it at working mag, then one just needs to load the GEOM into a shapely object and the bounding box is readily available?

jacksonjacobs1:
Due to the predictable nature of tiles, we removed the GEOM field from the tile table

choosehappy:
> Due to the predictable nature of tiles, we removed the GEOM field from the tile table

g-damn we're clever..... i guess then this line of my code isn't going to work :  )

https://github.com/choosehappy/QuickAnnotator/blob/a432fd8a5ec50038d1a2357d8da9dcf48e807996/quickannotator/dl/utils.py#L50

jacksonjacobs1:
Does this approach for getting and using tileSpace objects make sense to you? It makes a lot more sense to me, especially when multiple operations need to be applied to the same coordinate system, e.g., 

```python
tilespace = get_tilespace(args['annotation_class_id'], args['image_id']) # in_work_mag is True by default
bboxes = [tilespace.get_bbox_for_tile(id) for id in tilespace.get_all_tile_ids_for_image()]
```

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -42,33 +49,75 @@ def insert_new_annotation(image_id, annotation_class_id, is_gt, tile_id, polygon
...
+        image_id (int): The ID of the image to retrieve.
+        annotation_class_id (int): The ID of the annotation class to use for scaling.
+    Returns:
+        Query: A query object for retrieving annotations scaled to the appropriate magnification.
+    """
+    
+    scale_factor = base_to_work_scaling_factor(image_id, annotation_class_id)
+    return get_annotation_query(model, scale_factor)
+
+def base_to_work_scaling_factor(image_id: int, annotation_class_id: int) -> float:
```
choosehappy, line 109: my only concern overall is with this function - here you're using large_image to do some computation and come up with a scaling factor - which is totally fine. *however* you then use that scaling factor in multiple other places, essentially "doing the math" yourself. since in the "heart" of this code, we still have to create the ts pointer - why not use the large_image functions to do our conversions? e.g. 

```
sizeXtargetmag, sizeYtargetmag= ts.getPointAtAnotherScale((ts.sizeX,ts.sizeY),targetScale={'magnification': self.magnification}, targetUnits='mag_pixels') 
```

which gets large_image to do the work for us. 95% of the time i think it'll work fine, maybe even 99%, but i do wonder if there are edge cases, which if its internal calculations produce different results than our "external" calculations, things won't line up (similar to why we didn't want to mix openslide /large_image) - if we always use the same library everywhere - we eliminate that possibility

quickannotator/api/v1/tile/helper.py
```diff
@@ -63,113 +65,222 @@ def upsert_tile(annotation_class_id: int, image_id: int, tile_id: int, seen: Til
...
-
-    # Flatten the mesh grid and calculate tile IDs
-    tile_ids = (rows * tiles_per_row + cols).flatten().tolist()
-
-    return tile_ids
-
-def point_to_tileid(tile_size: int, image_width: int, image_height: int, x: int, y: int) -> int:
-    if not (0 <= x < image_width and 0 <= y < image_height):
-        raise ValueError(f"Point {x}, {y} is out of image dimensions (0, 0, {image_width}, {image_height})")
+class TileSpace:
```
choosehappy, line 68: love this idea - very neat and organized! i wonder however if it makes sense *here* or as helper functions on the Tile model itself? like Tile.get_tile_ids_within_bbox or Tile.utils.get_tile_ids_within_bbox? since all of that stuff is somehow related to that table? or perhaps i'm over thinking?

choosehappy (commented):
Amazing work - really nice and clean. some super mature code style in here. left a few comments for discussion before approving

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -42,33 +49,75 @@ def insert_new_annotation(image_id, annotation_class_id, is_gt, tile_id, polygon
...
+            func.ST_AsGeoJSON(model.polygon).label('polygon'),
+            model.area,
+            model.custom_metrics,
+            model.datetime
+        )
+    else:
+        query = db_session.query(
+            model.id,
+            model.tile_id,
+            func.ST_AsGeoJSON(func.ST_Scale(model.centroid, scale_factor, scale_factor)).label('centroid'),
```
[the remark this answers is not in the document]
jacksonjacobs1: Not yet, will try this and modify accordingly.

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: Yes this looks great - however I think we still need the scaling factor to pass into our db queries for getting annotations, right? It doesn't look like large_image has a method for getting the scale factor between 2 magnifications.

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: I see what you mean - I think there's a conceptual difference because when you initialize a tile:
```
models.Tile(annotation_class_id=2, image_id=1) 
```

you get a single instance of a tile, whereas the TileSpace methods are instead bound to a coordinate system and can e.g., get multiple tile ids. It wouldn't make sense to have to initialize a tile object to get other tile ids.

Alternatively you could move all the TileSpace methods into Tile as static methods, but then you don't get the benefit of storing the annotation_class_id and image_id in the object itself - these would once again become method params

Maybe what makes the most sense is to make get_tilespace a static method of Tile, still returning a TileSpace object?

Then you'd call
```python
Tile.get_tilespace(image_id, annotation_class_id, True).get_all_tile_ids_for_image()
```

choosehappy: yes, this latter one is what i was thinking - for exactly the reason you specified - still want to return an instance of the class which has those parameters set so that it can be freely passed around. how do you feel about it?

you can tell from this line of code that it then "Feels" to tightly couple the tilespace with the tile - and puts all the code in the same place. if you were a e.g., 3rd party developer, would this enhance your understanding or confuse it?

quickannotator/api/v1/utils/shared_crud.py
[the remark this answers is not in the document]
choosehappy: honestly- in thinking about it again - even when looking at your benchmarks, i think we saw a .01ms penalty for scaling an object.  in terms of hours of debugging time it would take to figure out an error - it seems like a no brainer to simplify the code ;-) the issue would be of course if there was an implementation/bug issue in one branch of the if-statement, we would be less likely to see it, and it may appear "Erratically" instead of consistently. given that i think it'll take only a couple of minutes to do the benchmark, i think we should do it anyway, however...even if there is a very very small (to be defined later) penalty, its probably worth to accept it from a code maitenance standpoint?

quickannotator/api/v1/utils/shared_crud.py
choosehappy: hmmm yes, you're correct - it doesn't look like there is a method for getting the scaling factor - i guess they would be reluctant to make such a method as well since it seems trivial, i.e., if you ask for the scaling factor between 40 and 10 - you're really just dividing the two values which is simple enough. i guess my concern is more about what happens at unexpected cases, e.g., 40x to 13x, like, are there specific (e.g., rounding to integer) type errors that will only occur if the factor is a float and not an int?

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Yes makes sense to me. I'll make this change.

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: I don't think so - all of our scaling methods accept a float, and I think large_image accepts unfriendly magnifications. 

Anyway, we should be guaranteed through form checking that the working mag is set to a friendly value (we will only give the user a dropdown to select from). We can also enforce that the basemag is friendly when the user uploads an image.

choosehappy: seems legit!

quickannotator/api/v1/tile/helper.py
choosehappy: as you wish : )

jacksonjacobs1: Ran into a problem adding get_tilespace to Tile causes circular imports because this method imports get_annotation_class_by_id and get_image_by_id from helper modules. The helper modules import the models module.

I think best practice is to keep get_tilespace in a helper module.

quickannotator/api/v1/tile/__init__.py
```diff
@@ -110,11 +110,11 @@ def get(self, args):
...
         """
         
-        tilespace = get_tilespace(args['annotation_class_id'], args['image_id'], in_work_mag=False)
-        tile_ids_in_bbox = tilespace.get_tile_ids_within_bbox(args['x1'], args['y1'], args['x2'], args['y2'])
+        tilespace = get_tilespace(image_id=args['image_id'], annotation_class_id=args['annotation_class_id'], in_work_mag=False)
+        tile_ids_in_bbox = tilespace.get_tile_ids_within_bbox((args['x1'], args['y1'], args['x2'], args['y2']))
         tile_ids_in_mask, _, _ = get_tile_ids_intersecting_mask(args['image_id'], args['annotation_class_id'], mask_dilation=1)
         ids = set(tile_ids_in_bbox) & set(tile_ids_in_mask)
-
+        tilespace.get_all_tile_ids_for_image()
```
choosehappy, line 117: does this line of code do anything? it doesn't seem to have its result stored?

quickannotator/api/v1/tile/helper.py
```diff
@@ -65,222 +60,46 @@ def upsert_tile(annotation_class_id: int, image_id: int, tile_id: int, seen: Til
...
-    r = base_to_work_scaling_factor(image_id, annotation_class_id)
-    if in_work_mag:
-        image_work_width = image.base_width * r
-        image_work_height = image.base_height * r
-        return TileSpace(annotation_class.work_tilesize, image_work_width, image_work_height)
-    else:   # return
-        base_tilesize = annotation_class.work_tilesize / r
-        return TileSpace(base_tilesize, image.base_width, image.base_height)
-
 # DEPRECATED
```
choosehappy, line 63: which part is now deprecated?

choosehappy (approved):
looks great! 2 minor things in comments

quickannotator/api/v1/tile/__init__.py
jacksonjacobs1: Nice catch - this must have been a typo. I've removed this line of code.

PR #84, opened 2025-03-12
Lasso functionality

### LINKED ISSUE #56: Annotation Page - lasso preds as gt
## Description
The user should be able to select certain predicted polygons and assign them as ground truth. This functionality will be accomplished via a lasso tool.

## Acceptance Criteria
- [ ] Clicking the lasso tool button enables the tool.
- [ ] When the user draws a polygon with the lasso tool enabled, pred polygons within the lasso are highlighted with white borders.
- [ ] When the user releases the mouse, the lasso tool is disabled, and the tool returns to the default mode (navigation).
- [ ] When the user releases the mouse, highlighted predicted polygons are committed as ground truth, are removed from the predictions table, and the relevant tile in the viewport is rerendered.

jacksonjacobs1:
Closes #56


quickannotator/api/v1/annotation/__init__.py
```diff
@@ -176,6 +184,25 @@ def get(self, args, image_id, annotation_class_id):
...
+        """     post multiple new annotations to the db.
+        
+        This method is primarily used for ground truth annotations. Predictions should only be saved by the model.
+        """
+        isgt = True
+        polygons = [shape(annotation['polygon']) for annotation in args]
+
+        annotations = bulk_insert_annotations(image_id, annotation_class_id, isgt, polygons, from_work_mag=False)
+        tileids = [ann.tile_id for ann in annotations]
+        bulk_upsert_tiles(annotation_class_id, image_id, tileids, hasgt=isgt)
```
choosehappy, line 201: i can't see from this review scope -- what is the purpose of this? i see insert and then and an upsert? what functionlaity is not completed in the first insert?

quickannotator/api/v1/annotation/__init__.py
```diff
@@ -185,6 +212,27 @@ def get(self, args, image_id, annotation_class_id, tile_id):
...
+@bp.route('/<int:image_id>/<int:annotation_class_id>/withinpoly')
+class AnnotationsWithinPolygon(MethodView):
+    @bp.arguments(GetAnnWithinPolyArgsSchema, location='json')
+    @bp.response(200, AnnRespSchema(many=True))
+    def post(self, args, image_id, annotation_class_id):
+        """     get all annotations within a polygon
+        """
+
+        # 1. Get all tiles intersecting the polygon
+        tiles_ids_within_mask, _, _ = get_tile_ids_intersecting_mask(image_id, annotation_class_id, constants.MASK_DILATION)
```
choosehappy, line 225: wow, lot of queries here -- can you walk me through the logic? i would have expected to be able to e.g., send the polygon to the database and perform the intersection there in the GEOM space. i suspect with the spatial index that would be extremely quick - on par with using the tile_id to pre-filter?

quickannotator/api/v1/tile/__init__.py
```diff
@@ -48,6 +48,13 @@ class SearchTileArgsSchema(Schema):
     x2 = fields.Float(required=True)
     y2 = fields.Float(required=True)
 
+class SearchTileByPolygonArgsSchema(Schema):
+    image_id = fields.Int(required=True)
+    annotation_class_id = fields.Int(required=True)
+    hasgt = fields.Bool(required=False)
+    include_placeholder_tiles = fields.Bool(required=False, default=False)    # A placeholder tile is a placeholder tile that has not yet been created in the database.
```
choosehappy, line 55: i thought we opted to eliminate the idea of placeholder tiles? if the tile needs to be returned, then we might as well create it and make it real

quickannotator/api/v1/tile/__init__.py
```diff
@@ -133,6 +140,36 @@ def get(self, args):
...
     
+@bp.route('/search/polygon')
+class TileSearchByPolygon(MethodView):
+    @bp.arguments(SearchTileByPolygonArgsSchema, location='json')
+    @bp.response(200, TileRespSchema(many=True))
+    def post(self, args):
+        """     get all Tiles within a polygon
+        """
+        tiles_in_polygon, _, _ = get_tile_ids_intersecting_polygons(args['image_id'], args['annotation_class_id'], [args['polygon']], mask_dilation=1)
+        tile_ids_in_mask, _, _ = get_tile_ids_intersecting_mask(args['image_id'], args['annotation_class_id'], mask_dilation=1)
```
choosehappy, line 151: what is the difference between this polygon and mask? isn't one of tehse supposed to be the issue mask, but that would mean the second parameter isn't set correctly?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -45,6 +48,38 @@ def insert_new_annotation(image_id, annotation_class_id, is_gt, tile_id, polygon
...
+    scale_factor = base_to_work_scaling_factor(image_id, annotation_class_id) if from_work_mag else 1.0
+    
+    new_annotations = []
+    for polygon in polygons:
+        scaled_polygon = polygon if from_work_mag else shapely.affinity.scale(polygon, xfact=scale_factor, yfact=scale_factor, origin=(0, 0))
+            
+        centroid = scaled_polygon.centroid
+        tile_id = tilespace.point_to_tileid(centroid.x, centroid.y)
+        new_annotations.append({
+            "image_id": None,  # Ensure correct value
```
choosehappy, line 64: should these be None?

choosehappy (changes requested):
Some questions/comments provided

quickannotator/api/v1/annotation/__init__.py
```diff
@@ -94,21 +103,20 @@ def post(self, args, image_id, annotation_class_id):
...
+        isgt = True
+
         # Get the tile id.
-        tilespace = get_tilespace(image_id, annotation_class_id, in_work_mag=True)  # Polygon is already scaled to the working magnification
+
         tile_id = tilespace.point_to_tileid(poly.centroid.x, poly.centroid.y)
-        
-        ann = insert_new_annotation(image_id, annotation_class_id, True, tile_id, poly)
-        
+        ann = insert_new_annotation(image_id, annotation_class_id, isgt, tile_id, poly)
```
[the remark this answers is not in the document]
jacksonjacobs1: I agree - this should happen internally.

Alternatively, thoughts on getting rid of insert_new_annotation entirely? This would reduce code redundancy now that we have a bulk annotation insertion function which already gets the tileids internally.

quickannotator/api/v1/annotation/__init__.py
```diff
@@ -94,21 +103,20 @@ def post(self, args, image_id, annotation_class_id):
...
+
         poly = scale(geom=shape(args['polygon']), xfact=scale_factor, yfact=scale_factor, origin=(0, 0))
-        
-        
+        isgt = True
+
         # Get the tile id.
-        tilespace = get_tilespace(image_id, annotation_class_id, in_work_mag=True)  # Polygon is already scaled to the working magnification
+
         tile_id = tilespace.point_to_tileid(poly.centroid.x, poly.centroid.y)
```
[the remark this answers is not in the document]
jacksonjacobs1: Okay, will do.

quickannotator/api/v1/annotation/__init__.py
jacksonjacobs1: Yeahhh... this should definitely be done in the database. I wasn't sure if sqlite would actually use the spatial index automatically, as we previously observed. That said, if we add a tileid filter to the spatial query, we can still get reasonable performance with sqlite while still using a db agnostic query.

I'm going to fill in the `get_annotations_within_polygons()` function below to perform the spatial query directly on the database. Sound good?

quickannotator/api/v1/annotation/helper.py
```diff
@@ -31,6 +31,16 @@ def get_annotations_for_tile(image_id: int, annotation_class_id: int, tile_id: i
...
 
+def get_annotations_for_tiles(image_id: int, annotation_class_id: int, tile_ids: List[int], is_gt: bool) -> List[models.Annotation]:
+    model: models.Annotation = create_dynamic_model(build_annotation_table_name(image_id, annotation_class_id, is_gt))
+    scale_factor = base_to_work_scaling_factor(image_id, annotation_class_id)
+    result: List[models.Annotation] = get_annotation_query(model, 1/scale_factor).filter(model.tile_id.in_(tile_ids)).all()
+    
+    return result
+
+def get_annotations_within_poly(image_id: int, annotation_class_id: int, is_gt: bool, polygon: geojson.Polygon) -> List[models.Annotation]:
+    pass
```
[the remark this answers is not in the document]
jacksonjacobs1: Yes will put the spatial query logic here.

quickannotator/api/v1/tile/__init__.py
jacksonjacobs1: Yes, I think we discussed in another thread just having this endpoint return tileids and letting the backend handle tile creation logic when the client calls an /infer endpoint on the tileid? Sound good?

quickannotator/api/v1/tile/__init__.py
jacksonjacobs1: This is set correctly. The `get_tile_ids_intersecting_mask` function knows the mask annotation class id implicitly (it's a global variable set to 1). The function needs to know the current annotation class id so that it can get the full set of tiles for the annotation class and perform the intersection with the mask.

I can add a docstring to the function to make this more clear

quickannotator/api/v1/tile/__init__.py
```diff
@@ -133,6 +140,36 @@ def get(self, args):
...
+        
+        tiles = query.all()
+
+        if args['include_placeholder_tiles']:
+            existing_ids = set([tile.tile_id for tile in tiles])
+            placeholder_tile_ids = ids - existing_ids
+            placeholder_tiles = [models.Tile(tile_id=tile_id, image_id=args['image_id'], annotation_class_id=args['annotation_class_id'], seen=TileStatus.UNSEEN, hasgt=False) for tile_id in placeholder_tile_ids]
+            
+            tiles.extend(placeholder_tiles)
+        return tiles, 200
```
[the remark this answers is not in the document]
jacksonjacobs1: Sure - I can make this change. see above comment about placeholder tiles

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: Yes, the `image_id` and `annotation_class_id` and `isgt` columns are placeholders which are not used by the rest of the application because they are redundant (contained in the table name).

I think we discussed keeping these columns in the Annotation model so that we don't need to modify the db schema if we decide to combine all annotation tables into a single (partitioned) table.

I left the following comment in the Annotation model:
```
# These columns are kept null intentionally because the information is stored in the table name. Consider eventually removing these columns.
```

That said, I would be okay with removing these columns if you think that's the right move - it won't have an impact on the rest of the application.

quickannotator/api/v1/annotation/__init__.py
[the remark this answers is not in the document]
choosehappy: i'm okay to remove it, but would rename the bulk one to be insert_new_annotation, and be able to submit either 1 annotation or multiple annotation. "bulk" doesn't sit well with the idea of inserting only 1 annotation, so its more of a naming thing.

btw- what is the difference between an "annotation" and a "polygon" - are these two distinct things or the same thing? i remember some function names having polygon and others having annotation?

quickannotator/api/v1/annotation/__init__.py
choosehappy: OoOO i see - one is for annotations and one is for tiles - sorry missed that

quickannotator/api/v1/tile/__init__.py
choosehappy: yup, that sounds familiar!

quickannotator/api/v1/annotation/__init__.py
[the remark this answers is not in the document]
jacksonjacobs1: Annotation refers to an item which exists in the database, and always follows the Annotation model.

Polygon is just a polygon, although its used interchangeably to describe geojson polygons, shapely polygons, WKT polygons, etc. I've tried to explicitly annotate the type in each case.

quickannotator/api/v1/annotation/__init__.py
choosehappy: yup!

quickannotator/api/v1/utils/shared_crud.py
choosehappy: rrrrrrrrrrrrright - thanks for reminding me. i think its up to you - at least for me, i would be confused by having them and would "accidently" set and/or use them because i saw them there. its probably easier to comment them out and then uncomment them later if needed, as opposed to potentially accidently using them?

jacksonjacobs1:
Notes: 
1. You'll have to rebuild the database since annotation_class_id, image_id, and isgt columns were commented out in the Annotation model
2. I've grouped all annotation CRUD into a class called `AnnotationStore`. This is less error prone because the class is initialized with a param `in_work_mag=True`, which automatically determines whether or not the functions accept & return polygons in the work magnification. Resulting API endpoints for getting & posting annotations are now much sleeker.

The pattern for inserting annotations is now:
```python
store = AnnotationStore(image_id, annotation_class_id, is_gt=True, in_work_mag=False)
anns = store.insert_annotations([shape(poly) for poly in polygons])
```
and the other methods have a similar syntax.

The AnnotationStore also has a static method `create_annotation_table()`. This gives us two ways to create the table:

```python
AnnotationStore.create_annotation_table(image_id, constants.MASK_CLASS_ID, is_gt=True)
```

```python
mask_store = AnnotationStore(image_id, constants.MASK_CLASS_ID, is_gt=True, create_table=True)
```

quickannotator/db/models.py
```diff
@@ -83,11 +84,11 @@ class Tile(Base):
...
 
     # columns
-    pred_status = Column(Integer, nullable=True, default=TileStatus.UNSEEN)
-    pred_datetime = Column(DateTime, nullable=True, server_default=func.now())
+    pred_status = Column(Integer, nullable=False, default=TileStatus.UNSEEN)
+    pred_datetime = Column(DateTime, nullable=True)
 
-    gt_counter = Column(Integer, nullable=True, default=0)
-    gt_datetime = Column(DateTime, nullable=True, server_default=func.now())
+    gt_counter = Column(Integer, nullable=False, default=0)
```
jacksonjacobs1, line 90: Did you intend for the gt_counter to be nullable? Normally I'd think the default state of a counter would be 0

quickannotator/db/models.py
```diff
@@ -83,11 +84,11 @@ class Tile(Base):
     tile_id = Column(Integer, nullable=False)
 
     # columns
-    pred_status = Column(Integer, nullable=True, default=TileStatus.UNSEEN)
-    pred_datetime = Column(DateTime, nullable=True, server_default=func.now())
+    pred_status = Column(Integer, nullable=False, default=TileStatus.UNSEEN)
```
jacksonjacobs1, line 87: pred_status should not be nullable since the default state is `UNSEEN`. Do you agree?

jacksonjacobs1:
This isn't ready for a full review yet, but I wanted to check that my updates to the Tile model schema are fine with you. Thoughts?

quickannotator/db/models.py
choosehappy: i did - my thinking was that for consistency, if gt_datetime is Null, then gt_counter should be null as well. 

having a gt_counter of "0" seems to suppose that the tile does indeed have a GT. i like, don't know what a counter =0 but a datetime = Null "means" -- it seems like a weird stage

quickannotator/db/models.py
choosehappy: UhhhhhhhHHHHhhhh...hmm....yes? yes. i think so. so this is basically to say that there is no "less active" state than UNSEEN? yea, it checks out

quickannotator/api/v1/annotation/__init__.py
```diff
@@ -89,119 +75,70 @@ class Annotation(MethodView):
...
         
         This method is primarily used for ground truth annotations. Predictions should only by saved by the model.
         """
-        # scale the polygon to the working magnification
-        scale_factor = base_to_work_scaling_factor(image_id, annotation_class_id)
-        tilespace = get_tilespace(image_id, annotation_class_id, in_work_mag=True)  # Polygon is expected to be scaled to the working magnification
-
-        poly = scale(geom=shape(args['polygon']), xfact=scale_factor, yfact=scale_factor, origin=(0, 0))
-        isgt = True
+        polygons: List[geojson.Polygon] = args['polygons']
```
choosehappy, line 92: i'm not 1000% sure that this would work if one sent just a polygon that wasn't "within" a list. like the difference between "poly" and "[poly]" - usually one supports both?

quickannotator/api/v1/tile/helper.py
```diff
@@ -201,9 +123,9 @@ def remote_compute_on_tile(annotation_class_id: int, image_id: int, tile_id: int
         # Process the tile (using shapely for example)
         bbox = tilespace.get_bbox_for_tile(tile_id)
         bbox_polygon = Polygon([(bbox[0], bbox[1]), (bbox[2], bbox[1]), (bbox[2], bbox[3]), (bbox[0], bbox[3])])
-        for _ in range(random.randint(20, 40)):
-            polygon = generate_random_circle_within_bbox(bbox_polygon, 100)
-            insert_new_annotation(image_id, annotation_class_id, is_gt=False, tile_id=tile_id, polygon=polygon)
+        polygons = [generate_random_circle_within_bbox(bbox_polygon, 100) for _ in range(random.randint(20, 40))]
```
choosehappy, line 126: i guess this is place holder code? : ) for sanity sake - can we mark it with "#TODO: [comment]" so that we can quickly grep for TODOs later on and make sure we don't miss anytihng?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -100,26 +45,15 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
-        query = db_session.query(
-            model.id,
-            model.tile_id,
-            func.ST_AsGeoJSON(func.ST_Scale(model.centroid, scale_factor, scale_factor)).label('centroid'),
-            func.ST_AsGeoJSON(func.ST_Scale(model.polygon, scale_factor, scale_factor)).label('polygon'),
-            model.area,
-            model.custom_metrics,
-            model.datetime
-        )
+    query = db_session.query(
```
choosehappy, line 48: beautiful, love this - super clean and sleek

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -137,3 +71,170 @@ def get_basemag_annotation_query(model, image_id, annotation_class_id):
...
+        table_name = build_annotation_table_name(image_id, annotation_class_id, is_gt=is_gt)
+        table = models.Annotation.__table__.to_metadata(Base.metadata, name=table_name)
+        Base.metadata.create_all(bind=db_session.bind, tables=[table])
+
+        return create_dynamic_model(table_name)
+
+
+
+    @staticmethod
+    def scale_polygon(polygon: BaseGeometry, scaling_factor: float) -> BaseGeometry:   # Added for safety - I've forgotten the origin param several times.
```
choosehappy, line 223: works for me

choosehappy (commented):
another question - are you running the tests before doing commits to ensure things still work as expected?

choosehappy:
> This isn't ready for a full review yet, but I wanted to check that my updates to the Tile model schema are fine with you. Thoughts?

did i provide feedback on what you were looking for?

quickannotator/api/v1/annotation/__init__.py
jacksonjacobs1: Yeah... this endpoint definitely only supports a list type at the moment. 

It looks like we could define a `@preload` method within endpoints which accept either lists or singletons for a particular parameter.
https://chatgpt.com/share/67d885a5-dcc0-8003-9b56-7f5f165265e9

This method would automatically wrap a polygon parameter in a list.

Is this what you had in mind, or did you want singleton/list to be passed further into the python code? (e.g., `AnnotationStore.get_annotations_for_tiles()` accepts either a list or a singleton)

quickannotator/db/models.py
jacksonjacobs1: Okay, gt_counter is nullable again with the default value set to null.

quickannotator/api/v1/annotation/__init__.py
choosehappy: interesting - didn't know about this preload functionality.  a few questions:

1) it looks like insert_annotations only takes a single polygon anyway? is that correct? would "going deeper" provide any additional functionality in terms speed?  i don't imagine so

2) what is the value of a preload? does it somehow change the documentation/swagger presnetation so that it says it accepts both list or a single polygon?

3) preload seems cool - without though i probably would have just used a single if statement between 90 and 91:

```python
poly = [poly] if not isinstance(poly,list) else poly
```

which seems less....hardcore ... than a preloader - wihch is why i wonder what additional functionality it provides?

jacksonjacobs1: `insert_annotations` takes a polygon list as input and performs a bulk insertion

I think preload is required because flask-smorest validates the arguments against the arguments schema before passing them into the method. A single polygon would fail during this validation step.

Unfortunately preload does not change the swagger docs.

Essentially we want to use `oneOf` in swagger:
https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/

Which doesn't seem to be supported in flask-smorest:
https://github.com/marshmallow-code/flask-smorest/issues/80

I'm okay to use preload, but it's a bit of a janky solution. I think it might be better to accept a list?

jacksonjacobs1:
> another question - are you running the tests before doing commits to ensure things still work as expected?

Yes - I've updated the tests wrt. the latest commit - they are all currently passing

quickannotator/api/v1/annotation/__init__.py
choosehappy: yea - i think we're going in the weeds here. okay to require a list as long as its clearly documented/obvious from the inputs

quickannotator/api/v1/annotation/helper.py
```diff
@@ -1,72 +1,44 @@
...
+# def annotations_within_bbox_spatial(table_name: str, x1: float, y1: float, x2: float, y2: float) -> List[models.Annotation]:
+#     stmt = text(f'''
+#         SELECT ROWID, AsGeoJSON(centroid) as centroid, area, AsGeoJSON(polygon) as polygon, custom_metrics, datetime, annotation_class_id
+#         FROM "{table_name}"
+#         WHERE "{table_name}".ROWID IN (
+#             SELECT ROWID
+#             FROM SpatialIndex
+#             WHERE f_table_name = "{table_name}"
+#             AND f_geometry_column = 'centroid'
+#             AND search_frame = BuildMbr({x1}, {y1}, {x2}, {y2})
```
choosehappy, line 26: btw - even if we at some point need to write sql code- i would still like to err on the side of using ORM, which should decrease the error rate. chatgpt does a good job it looks like 

```python
from sqlalchemy.orm import Session
from sqlalchemy import select, text
from sqlalchemy.sql import func
from geoalchemy2.functions import ST_AsGeoJSON
from your_models import YourTable, SpatialIndex  # Replace with actual model names

def query_polygons(session: Session, table_name: str, x1: float, y1: float, x2: float, y2: float):
    subquery = (
        select(SpatialIndex.ROWID)
        .where(
            SpatialIndex.f_table_name == table_name,
            SpatialIndex.f_geometry_column == 'centroid',
            SpatialIndex.search_frame == func.BuildMbr(x1, y1, x2, y2)
        )
    )

    query = (
        session.query(
            YourTable.ROWID,
            ST_AsGeoJSON(YourTable.centroid).label("centroid"),
            YourTable.area,
            ST_AsGeoJSON(YourTable.polygon).label("polygon"),
            YourTable.custom_metrics,
            YourTable.datetime,
            YourTable.annotation_class_id,
        )
        .filter(YourTable.ROWID.in_(subquery))
    )

    return session.execute(query).fetchall()


```

quickannotator/api/v1/tile/__init__.py
```diff
@@ -25,134 +24,125 @@ class PredictTileRespSchema(Schema):
...
         """     delete a Tile
-
         """
-
+        tile = db_session.query(models.Tile).filter_by(
+            image_id=image_id,
+            annotation_class_id=annotation_class_id,
+            tile_id=args['tile_id']
+        ).first()
+        if tile:
```
choosehappy, line 77: sorry - why is this a query + delete instead of just a delete?

quickannotator/api/v1/tile/__init__.py
```diff
@@ -25,134 +24,125 @@ class PredictTileRespSchema(Schema):
...
-            
-            tiles.extend(placeholder_tiles)
-        return tiles, 200
+
+        # Filter by tiles which have ground truths saved. This limits the number of tiles we have to consider for rendering.
+        if args['hasgt']:
+            ids = db_session.query(models.Tile.tile_id).filter(
+                models.Tile.annotation_class_id == annotation_class_id,
+                models.Tile.image_id == image_id,
+                models.Tile.tile_id.in_(ids),
```
choosehappy, line 110: tile_id.in_(ids) <--- is this correct? is the second id the row id or the tile_id?

quickannotator/api/v1/tile/__init__.py
```diff
@@ -25,134 +24,125 @@ class PredictTileRespSchema(Schema):
...
+                models.Tile.gt_datetime.isnot(None)
+            ).all()
+            ids = [id[0] for id in ids]
+        return {"tileids": ids}, 200
```
choosehappy, line 114: i think we need to be very careful here - because that Tiles table also has an id column, doesn't it? ideally, variable names should be 1:1 with the names of the columns in the DB, at minimum they shouldn't conflict with another columns name ;-)

quickannotator/api/v1/tile/__init__.py
```diff
@@ -25,134 +24,125 @@ class PredictTileRespSchema(Schema):
...
-    @bp.response(201, PredictTileRespSchema)
-    def post(self, args):
-        """     predict tiles for a given image & class
-        """
-        upsert_tile(args['annotation_class_id'], args['image_id'], args['tile_id'], seen=TileStatus.PROCESSING)
-
-        object_ref = compute_on_tile(args['annotation_class_id'], args['image_id'], tile_id=args['tile_id'], sleep_time=5)
-        return {'object_ref': object_ref}, 201
-        
+        return {"tileids": [tile_id]}, 200
```
choosehappy, line 147: this naming is getting a bit wild .. in this screen alone i see tileids, tile_id, and ids and i think they're all actually instances of the same information?

quickannotator/api/v1/tile/helper.py
```diff
@@ -76,37 +38,53 @@ def tile_intersects_mask_shapely(image_id: int, annotatation_class_id: int, tile
...
+        tuple: A tuple containing:
+            - tile_ids (list[int]): A list of tile IDs that intersect with the polygons.
+            - mask (np.ndarray): The binary mask image with filled polygons.
+            - processed_polygons (list[np.ndarray]): A list of processed polygons with scaled coordinates.
+    """
+
+    image = get_image_by_id(image_id)
+    base_tilesize = get_annotation_class_by_id(annotation_class_id).work_tilesize / base_to_work_scaling_factor(image_id=image_id, annotation_class_id=annotation_class_id)
+    processed_polygons = []
+    scale_factor = 1 / base_tilesize
```
choosehappy, line 74: do you want to put the scaling in the DB instead of in the backend?

quickannotator/api/v1/tile/helper.py
```diff
@@ -138,24 +115,23 @@ def remote_compute_on_tile(annotation_class_id: int, image_id: int, tile_id: int
     with get_session() as db_session:
         # Example: load the tile and process
         # breakpoint()
-        tile = get_tile(annotation_class_id, image_id, tile_id)  # Replace with your actual function to get the tile
+        tile = get_tile(image_id, annotation_class_id, tile_id)  # Replace with your actual function to get the tile
```
choosehappy, line 118: not sure what this comment means?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -65,40 +45,181 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+    # CREATE
+    # TODO: make this function applicable to predicted annotations, not just ground truths.
+    def insert_annotations(self, polygons: List[BaseGeometry]) -> List[models.Annotation]:
+        # in_work_mag is true because we expect the polygons are scaled at this point.
+        tilespace = get_tilespace(self.image_id, self.annotation_class_id, in_work_mag=True)
+        new_annotations = []
+        for polygon in polygons:
+            scaled_polygon = self.scale_polygon(polygon, self.scaling_factor)
+            tile_id = tilespace.point_to_tileid(scaled_polygon.centroid.x, scaled_polygon.centroid.y)
+
```
choosehappy, line 141: do you want to do the scaling in the DB while inserting?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -65,40 +45,181 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+            annotation.custom_metrics = compute_custom_metrics()
+            annotation.datetime = datetime.now()
+            db_session.commit()
+            return self.get_annotation_by_id(annotation_id)  # Get the updated annotation by id
+
+        return None  # Handle case where annotation_id is not found
+
+
+    # DELETE
+    def delete_annotation(self, annotation_id: int):
```
choosehappy, line 204: i think there is a .delete function that supports the filter? : https://docs.sqlalchemy.org/en/20/tutorial/data_update.html#the-delete-sql-expression-construct

quickannotator/db/models.py
```diff
@@ -136,7 +137,7 @@ class Notification(Base):
     message_type = Column(Integer, nullable=False)
     is_read = Column(Boolean, nullable=False)
     message = Column(Text, nullable=False)
-    datetime = Column(DateTime, server_default=func.now())
+    datetime = Column(DateTime)
```
choosehappy, line 140: i guess this was the reason for using now()? tit was possible to set the default? i guess one can't use e.g., datetime.now as the default?

choosehappy (changes requested):
looking good - a few open questions from my side

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: These polygons are coming from the client and not the database, so the scaling has to happen in the backend. Do you see another way?

quickannotator/api/v1/tile/helper.py
jacksonjacobs1: Left over from chatgpt. I've added a TODO to remove this method.

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: I don't think we can, because the area + custom metrics need to be computed on the scaled polygon

quickannotator/db/models.py
jacksonjacobs1: Yes. I guess we can still use func now in model schema where we don't need e.g., bulk insertion, but I don't personally see the disadvantage of using serverside `datetime.now()`. Do you have anything in mind?

choosehappy: 1. sounds right
2. not sure - it should default into the correct datetime, so it may work out? probably worth either testing.....or just set explicitly ;-)


what do you think?

choosehappy (approved):

quickannotator/db/models.py
jacksonjacobs1: Since bulk insertions bypass the ORM, they won't use default values specified in the model schema.

Anyways, I've gone and set the default to datetime.now in the respective models, and removed the explicit datetime.now() calls from e.g., add_project()

PR #85, opened 2025-04-02
Dl integration

jacksonjacobs1:
Initial dl component integration is complete.

### DIFF
--- a/quickannotator/api/v1/utils/shared_crud.py
+++ b/quickannotator/api/v1/utils/shared_crud.py
@@ -232,27 +232,24 @@ class AnnotationStore:
         if len(polygons) == 0:
             return []
         
-        if isinstance(tile_ids, int):
-            tile_ids = [tile_ids] * len(polygons)
-        
-        if tile_ids is not None and len(polygons) != len(tile_ids):
-            raise ValueError("The lengths of polygons and tile_ids must match if tile_ids is a list.")
+        if not isinstance(tile_ids, list):
+            tile_ids = [tile_ids] * len(polygons)   # By default produce a list of None values.
+
+        if len(polygons) != len(tile_ids):
+            raise ValueError("The lengths of polygons and tile_ids must match.")
 
         # in_work_mag is true because we expect the polygons are scaled at this point.
         tilespace = get_tilespace(self.image_id, self.annotation_class_id, in_work_mag=True)
         new_annotations = []
 
-        for i, polygon in enumerate(polygons):
+        for polygon, tile_id in zip(polygons, tile_ids):
             scaled_polygon = self.scale_polygon(polygon, self.scaling_factor)
 
-            # Determine tile_id based on provided parameters or calculate it
-            if tile_ids is not None:
-                current_tile_id = tile_ids[i]
-            else:
-                current_tile_id = tilespace.point_to_tileid(scaled_polygon.centroid.x, scaled_polygon.centroid.y)
+            if tile_id is None:
+                tile_id = tilespace.point_to_tileid(scaled_polygon.centroid.x, scaled_polygon.centroid.y)
 
             new_annotations.append({
-                "tile_id": current_tile_id,  # Ensure correct value
+                "tile_id": tile_id,  # Ensure correct value
                 "centroid": scaled_polygon.centroid.wkt,
                 "polygon": scaled_polygon.wkt,
                 "area": scaled_polygon.area,


quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -58,44 +59,132 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
 
     return query
 
-def upsert_tiles(image_id: int, annotation_class_id: int, tile_ids: List[int], pred_status: TileStatus = None):
+def reset_all_PROCESSING_tiles(annotation_class_id: int):
+    stmt = sqlalchemy.update(models.Tile).where(
+        models.Tile.annotation_class_id == annotation_class_id
+    ).values(
```
choosehappy, line 65: is this correct? i don't see a constraint for this to operate only on PROCESSING tiles? what about e.g., DONE? this seems like it would reset them as well?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -58,44 +59,132 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+                    (models.Tile.pred_datetime <= expiration_thresh), pred_status
+                ),
+            else_=models.Tile.pred_status,
+            ),
+            'pred_datetime': sqlalchemy.case(
+                (models.Tile.pred_status == TileStatus.UNSEEN, datetime.now()),
+                (
+                    (models.Tile.pred_status == TileStatus.DONEPROCESSING) & 
+                    (models.Tile.pred_datetime <= expiration_thresh), datetime.now()
+                ),
```
choosehappy, line 165: this makes me feel a bit uncomfortable because it'd take me a lot of brain power to parse and check : ) are you sure this is (a) correct and (b) the easiest way to do this?

quickannotator/db/annotation_crud.py
choosehappy, line 1: empty file?

quickannotator/dl/training.py
```diff
@@ -122,7 +123,7 @@ def train_pred_loop(config):
 
     running_loss = []
     
-    writer = SummaryWriter(log_dir=f"/tmp/{classid}/{datetime.datetime.now().strftime('%b%d_%H-%M-%S')}")
+    writer = SummaryWriter(log_dir=f"/tmp/{annotation_class_id}/{datetime.datetime.now().strftime('%b%d_%H-%M-%S')}")
```
choosehappy, line 126: these should go somewhere in the project directory and not /tmp

quickannotator/dl/training.py
```diff
@@ -98,8 +99,8 @@ def train_pred_loop(config):
     
     scaler = torch.amp.GradScaler("cuda")
 
-    print(summary(model, (3, tile_size, tile_size))) #TODO: log this
-
+    # print(summary(model, (3, tile_size, tile_size))) #TODO: log this
+    # TODO: why does the above line produce an error: RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same
```
choosehappy, line 103: strange - this would seem to mean that the function is creating a temporary tile on the GPU instead of in memory  - i don't have this error locally... if this error persists, i think you can add "device ='cpu'" to the call to force all types to be .Float and not .cuda.float

choosehappy (changes requested):
great first pass !

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: 1. Need on_conflict_do_update to keep the operation atomic
2. We didn't use case logic originally. The previous approach used a `where` clause to filter by UNSEEN or stale DONEPROCESSING. However, the previous approach only returned tiles which satisfied the update condition, rather than returning the tile regardless of whether or not it was updated.

That said, the current method is not very readable.

I've just pushed an update which should be more readable and has less redundant code. Thoughts?

quickannotator/api/v1/utils/shared_crud.py
jacksonjacobs1: You're right, this is incorrect.

Just pushed a fix + tests.

Note that this method isn't used yet but will be used if/when we implement a solution for perpetually PROCESSING tiles

quickannotator/db/annotation_crud.py
jacksonjacobs1: The project structure has gotten a little out of hand - I sought to separate database crud functions from the API helpers since I was running into some circular import errors.

The repository has since matured, i.e., functions are more modular now and the relationships between modules has been simplified, reducing the likelihood of circular imports, but the structure of the repository has gotten a little loose and should be tightened up in a PR dedicated to refactoring.

In the short term I've deleted this file and will write a feature request to pin down our final project structure.

choosehappy: understood, seems reasonable

quickannotator/api/v1/utils/shared_crud.py
choosehappy: ohhh yes - much more attractive and easy to parse, thanks

quickannotator/dl/training.py
jacksonjacobs1: Okay. I'm going to set up the output directory structure in the export functionality PR (not yet opened).

I've noted this request.

jacksonjacobs1:
I think this PR is ready to be merged. @choosehappy  do you agree?

quickannotator/api/v1/utils/shared_crud.py
```diff
@@ -131,32 +226,44 @@ def __init__(self, image_id: int, annotation_class_id: int, is_gt: bool, in_work
...
         new_annotations = []
-        for polygon in polygons:
+
+        for i, polygon in enumerate(polygons):
             scaled_polygon = self.scale_polygon(polygon, self.scaling_factor)
-            tile_id = tilespace.point_to_tileid(scaled_polygon.centroid.x, scaled_polygon.centroid.y)
+
+            # Determine tile_id based on provided parameters or calculate it
+            if tile_ids is not None:
+                current_tile_id = tile_ids[i]
```
choosehappy, line 250: this freaks me out a bit - this is not pythonic - a normal way would be to use a 
```python
for current_tile_id, polygon in zip(polygon,tile_ids):
```

which removes the chances for *a lot* of bugs. is that pattern applicable here?

jacksonjacobs1: Done. Looks much better now!

choosehappy (approved):
hot

Issue #86, opened 2025-04-03
[FEATURE] Add logging


jacksonjacobs1:
## Description

Implement logging functionality using Python’s built-in `logging` library, with logs being stored in a database using SQLAlchemy. This feature will allow users to track events, errors, and application behavior systematically. Logging to a database will enable easier querying, persistence, and analysis of log data.

Add a page for visualizing logs.

## Preconditions

- The application must have SQLAlchemy configured and connected to a database.
- The logging module should be available in the Python environment.

## Acceptance Criteria

- [ ] The feature enables logging of messages at different levels (DEBUG, INFO, WARNING, ERROR, CRITICAL).
- [ ] Logs are stored persistently in a database using SQLAlchemy.
- [ ] The logging implementation does not introduce significant performance overhead.
- [ ] Unit tests verify correct logging behavior and database integration.
- [ ] A grafana instance can be used for visualizing logs.

## Technical Notes

- Utilize Python’s `logging` module to handle log message creation.
- Implement a custom `logging.Handler` to store logs in the database.
- Define a SQLAlchemy model to represent log records in the database.

Consider the following schema:
```
id INTEGER PRIMARY KEY AUTOINCREMENT,
timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
level TEXT,
message TEXT

```

PR #88, opened 2025-04-04
Add logging

jacksonjacobs1:
Loggers can be initialized and used like so:
```python
from quickannotator.db.logging import init_logger
import logging

init_logger('logger_name')

logger = logging.getLogger('logger_name')
logger.info("test")
```

Logs are written to the `log` table:
```python
class Log(Base):

    __tablename__ = 'log'

    # primary key
    id = Column(Integer, primary_key=True, autoincrement=True)

    # columns
    name = Column(Text, nullable=False)
    timestamp = Column(DateTime, nullable=False)
    level = Column(Enum('DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'), nullable=False, default='INFO')
    message = Column(Text, nullable=False, default="")
```

And logs can be viewed from grafana.

image


jacksonjacobs1:
closes #86

quickannotator/db/logging.py
```diff
@@ -0,0 +1,27 @@
...
+                level=record.levelname,
+                message=self.format(record)
+            )
+
+            db_session.add(log_entry)
+
+
+def init_logger(logger_name) -> logging.Logger:
+    logger = logging.getLogger(logger_name)
+    logger.setLevel(logging.DEBUG)
```
choosehappy, line 21: for sanity sake, i think default level in most places is info?

choosehappy (changes requested):

quickannotator/db/logging.py
jacksonjacobs1: Sure, pushed this in on Friday.

I think this PR is ready to be merged unless you have any other requests?

I can also add documentation for setting up the grafana instance? The setup is very easy and only needs to be done once since it runs in a separate docker container.

choosehappy: okay for merge - i think everything was looking good

> I can also add documentation for setting up the grafana instance? The setup is very easy and only needs to be done once since it runs in a separate docker container

yes, probably will need this for the other developers (me, nan, etc)

PR #89, opened 2025-04-08
QuickAnnotator supports postGIS & spatialite

jacksonjacobs1:
QuickAnnotator is now database agnostic between spatialite and postgis.

Upsert_tiles() conditionally calls the dialect-specific insert statement depending on the engine dialect.

Additionally, here we release a grafana dashboard for viewing logs compatible with the postgres database.


choosehappy:
looks good, thanks!

jacksonjacobs1:
I've added TileStore & TileStoreFactory patterns.

choosehappy:
> 
> I would recommend waiting to review until `dl-integration` is merged into `v2.0`

you got it boss - let me know

quickannotator/api/v1/image/routes.py
```diff
@@ -0,0 +1,118 @@
...
+        return result, 200
+
+#################################################################################
+@bp.route('/<int:image_id>/<int:file_type>/file', endpoint="file")
+class ImageFile(MethodView):
+    def get(self, image_id, file_type):
+        """     returns an Image file   """
+        result = db_session.query(db_models.Image).filter(db_models.Image.id == image_id).first()
+
+        if file_type == 1:  # image file
```
choosehappy, line 69: same - enums are prefered here

quickannotator/api/v1/image/routes.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+
+        return 201
+
+    def delete(self, image_id, file_type):
+        """     delete an Image file   """
+
+        result = db_session.query(db_models.Image).filter(db_models.Image.id == image_id).first()
+
+        if file_type == 1:  # image file
```
choosehappy, line 89: same - enum

quickannotator/api/v1/image/routes.py
```diff
@@ -0,0 +1,118 @@
...
+            # TODO implement thumbnail file deletion
+            pass
+        return 204
+
+#################################################################################
+
+@bp.route('/<int:image_id>/patch_file/<int:level>/<int:col>_<int:row>.<string:file_format>', endpoint="patch")
+class PatchFile(MethodView):
+    def get(self, image_id, level, col, row, file_format):
+        """     returns a patch file   """
```
choosehappy, line 102: is file_format used in this function?

quickannotator/api/v1/tile/models.py
```diff
@@ -0,0 +1,48 @@
+from marshmallow import fields, Schema
+from marshmallow_sqlalchemy import SQLAlchemyAutoSchema
+import quickannotator.db.models as db_models
+
+
+class TileRespSchema(SQLAlchemyAutoSchema):
+    class Meta:
+        model = db_models.Tile
+        include_fk = True
```
choosehappy, line 9: what is fk?

choosehappy: foreign key?

quickannotator/api/v1/tile/routes.py
```diff
@@ -0,0 +1,124 @@
...
+        tilespace = get_tilespace(image_id=image_id, annotation_class_id=annotation_class_id, in_work_mag=False)
+        bbox = tilespace.get_bbox_for_tile(args['tile_id'])
+        geojson_polygon = {
+            "type": "Polygon",
+            "coordinates": [[
+                [bbox[0], bbox[1]],
+                [bbox[2], bbox[1]],
+                [bbox[2], bbox[3]],
+                [bbox[0], bbox[3]],
+                [bbox[0], bbox[1]]
```
choosehappy, line 70: interetsing - does the polygon need to be explicitly closed? i would not have thought to do that

quickannotator/config/config.ini
```diff
@@ -7,6 +7,7 @@ version = v1
 
 [sqlalchemy]
 database_uri = sqlite:////opt/QuickAnnotator/quickannotator/instance/quickannotator.db
+; database_uri = postgresql+psycopg2://YOURUSERNAME:YOURPASSWORD@postgis:5432/DATABASENAME ; NOTE: Use this URI for PostgreSQL. Replace with your credentials
```
choosehappy, line 10: is ";" a comment? i preferentially use "#"

quickannotator/db/crud/tile.py
```diff
@@ -0,0 +1,284 @@
...
+                    'annotation_class_id': annotation_class_id,
+                    'image_id': image_id,
+                    'tile_id': tile_id,
+                    **insert_fields
+                }
+                for tile_id in tile_ids
+            ]
+        ).on_conflict_do_update(
+            index_elements=['annotation_class_id', 'image_id', 'tile_id'],
+            set_=update_fields
```
choosehappy, line 60: does set have an underscore? set_?

quickannotator/db/crud/tile.py
```diff
@@ -0,0 +1,284 @@
...
+
+        
+class SQLiteTileStore(TileStore):
+    def __init__(self):
+        super().__init__(insert_method=sqlite_insert)
+
+
+class TileStoreFactory():
+    @staticmethod
+    def get_tilestore() -> TileStore:
```
choosehappy, line 274: nnnnnnnnice

choosehappy (approved):
really solid work, nice job - some minor comments for consideration but good to go from my side

quickannotator/api/v1/tile/models.py
jacksonjacobs1: Yes, this is the pattern for configuring the SqlAlchemyAutoSchema class, and here I specify to include foreign keys in the output. The foreign keys are annotation_class_id and image_id, which are needed in combination with tile_id to access the idx_annotation_class_image_tile index on Tile.

Here's the documentation for include_fk
https://marshmallow-sqlalchemy.readthedocs.io/en/latest/api_reference.html#marshmallow_sqlalchemy.SQLAlchemyAutoSchemaOpts

quickannotator/api/v1/tile/routes.py
jacksonjacobs1: Yes, all geojson polygons must be closed.

quickannotator/config/config.ini
jacksonjacobs1: The comment shortcut in vscode (ctrl+/) automatically uses ; for config.ini, so I'm inclined to use that character. Thoughts?

quickannotator/db/crud/tile.py
jacksonjacobs1: Yes, here's the postgis docs, and the interface is identical for the spatialite dialect insert method

https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.Insert.on_conflict_do_update.params.set_

jacksonjacobs1:
I've implemented all feedback, ran tests and tested functionality, so this is ready to be merged into v2.0

quickannotator/db/crud/tile.py
jacksonjacobs1: Looked it up and it's because "set" is a reserved keyword in python.

quickannotator/api/v1/annotation_class/routes.py
```diff
@@ -0,0 +1,84 @@
...
+    @bp.arguments(server_models.PostAnnClassArgsSchema, location='query')
+    @bp.response(200, description="AnnotationClass created")
+    def post(self, args):
+        """     create a new AnnotationClass   """
+
+        annotation_class = db_models.AnnotationClass(project_id=args['project_id'],
+                                          name=args['name'],
+                                          color=args['color'],
+                                          work_mag=args['work_mag'],
+                                          work_tilesize=2048
```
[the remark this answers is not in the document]
choosehappy: hmm...at minimum i think this value should appear in the config.ini, so that one doesn't have to e.g., hunt through an entire code base looking for it? as well, i guess i'm surprised that if this is fixed it is only fixed in this one location, or are there multiple places in the code which have the value of 2048 explicitly set (making it even more painful to hunt/change it in every location)

> Do you think I should add an option for tilesize in the annotation_class creation form?

i don't think so, this seems too "geeky" to have to burden our users with it - it seems more like something the IT person setting up this tool should be able to change since it is really critically connected with compute capability and not somehow the _actual_ functionality of the system - is that right?

quickannotator/db/crud/tile.py
choosehappy: yea - that makes sense....however if i were in their position i would likely think of another word to use instead of set, to avoid both hitting into a reserve word AND this weird underscore pattern (which i don't think i've seen anywhere else?)

quickannotator/config/config.ini
choosehappy: i asked chatGPT:


> In .ini configuration files, both ; and # are commonly used for comments, but the most widely accepted convention is to use the ; symbol. This is because the INI format was originally designed with semicolons in mind for comments. However, # is also supported by many parsers and is seen in some configurations, especially in environments where other types of configuration files (like shell scripts or certain other file formats) also use # for comments.
>
> So, while both work, it's safest to use ; for maximum compatibility with the INI standard. 


hard to argue with that...

PR #94, opened 2025-05-07
Class management

jacksonjacobs1:
image

image

image


jacksonjacobs1:
One problem that became clear in this PR is that the underlying DL integration is not yet stable.

When switching from one class to another, the `/startproc` endpoint is called and the HTTP connection is held open perpetually. Eventually this leads to unresponsive behavior. The root cause is that the startproc endpoint calls a remote function which interacts with the business layer: 

```python
date = ray.get(actor.getProcRunningSince.remote())
```

this function call sometimes never completes, but the reason for this is unclear and requires future work.

As this problem is more closely related to DL integration than class management, it should be addressed in a different PR.

jacksonjacobs1:
One possibility is this:

 If you call `start_dlproc.remote()` multiple times and the actor has long-running or blocking tasks (like deep learning inference), the actor’s internal task queue fills up. When you then call something like `getProcRunningSince.remote()`, that method call may sit queued behind long-running tasks, effectively hanging from your perspective.

quickannotator/db/colors.py
```diff
@@ -0,0 +1,21 @@
...
+
+        if color_palette_name not in constants.ANNOTATION_CLASS_COLOR_PALETTES:
+            raise KeyError(f"Color palette '{color_palette_name}' does not exist in constants.COLOR_PALETTES.")
+        
+        self.color_list = constants.ANNOTATION_CLASS_COLOR_PALETTES[color_palette_name]
+        self.project_id = project_id
+
+    def get_unused_color(self):
+        annotation_classes = search_annotation_class_by_project_id(self.project_id)
+        used_colors = {str(ac.color) for ac in annotation_classes}
```
choosehappy, line 16: only minor comment looking at this section - its not obvious to me if this is always going to be determinsitic or not, especially since "set" makes no gaurantee about ordering?  its not a requirement, but its a "nice to have" that e.g., the first class is always color C1, the second class is always C2, etc - so that when going interproject there is some constancy

quickannotator/db/crud/annotation.py
```diff
@@ -75,11 +72,15 @@ def __init__(self, image_id: int, annotation_class_id: int, is_gt: bool, in_work
...
 
-        if create_table:
-            model = self.create_annotation_table(image_id, annotation_class_id, is_gt)
-            self.model = model
+        table_name = build_annotation_table_name(image_id, annotation_class_id, is_gt=is_gt)
+        if table_exists(table_name):
+            self.model = create_dynamic_model(table_name)
         else:
-            self.model = create_dynamic_model(build_annotation_table_name(image_id, annotation_class_id, is_gt=is_gt))
+            if create_table:
```
choosehappy, line 79: out of curiosity - why not create it when its accessed? its a lightweight operation, and i'm guessing 99.9% of the time people use the "get"  they would set create = True?

choosehappy (approved):
Looks great - 2 minor questions

quickannotator/db/crud/annotation.py
jacksonjacobs1: code redundancy - this would require that the table checking & creation logic be moved to the top of each method.

I guess this logic could be performed in a function called `create_ifn_exists`, or even a decorator.

Thoughts?

choosehappy: my question is more from an API user perspective - should they care/be aware of if the table is exists and/or should they have the requirement to dictate when the table is created? i.e., have we delegated too much unwarranted responsbility to them?

jacksonjacobs1: I see what you mean. I guess we can just create the table if it doesn't exist. There's no need for the user to know the inner db interactions of AnnotationStore.

That said, I think we should add protections for the case where AnnotationStore is instantiated for a nonexistent annotation_class_id or image_id. In this case the __init__() function should raise an error.

Thoughts?

choosehappy: yes, that makes sense to me  - are you happy with these conclusions then?

PR #95, opened 2025-05-08
Viewport touchups

jacksonjacobs1:
Added mouse coordinates and scale indicator to the screen.

image

>**Important**
The scale has not been verified for correctness. In Fact it looks incorrect to me. I searched the HistomicsUI repository but was unable to identify the bug in my map/widget initialization which might be causing this.

### DIFF
--- a/quickannotator/client/src/components/legend.tsx
+++ b/quickannotator/client/src/components/legend.tsx
@@ -1,0 +1,34 @@
+import React, { useState } from "react";
+import Card from "react-bootstrap/Card";
+
+interface Props {
+    mouseCoords: number[];
+}
+
+const Legend = React.memo((props: Props) => {
+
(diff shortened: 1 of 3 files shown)


choosehappy:
just a quick question -- where/how do you calculate the bottom left's microns? i guess i would have expected something like a backend call to get the base microns and some constant conversion from that factor?

jacksonjacobs1:
You hit the nail on the head - this wasn't happening anywhere.

I would have liked to copy the pattern used in HistomicsUI directly but was unable to find it.

Instead I just set up an additional endpoint to return image metadata. Currently we just return mpp, which is then used by the viewport to set the scale.

Visual verification that this is working correctly:
image

choosehappy (approved):

PR #97, opened 2025-05-12
Export functionality

jacksonjacobs1:


quickannotator/api/v1/annotation/routes.py
```diff
@@ -129,4 +139,78 @@ def post(self, args):
...
+    @bp.arguments(server_models.DownloadAnnsArgsSchema, location='json')
+    def post(self, args):
+        """ Export annotations for multiple images and annotation classes as a TAR file with progress updates """
+
+        image_ids = np.array(args['image_ids'])
+        annotation_class_ids = np.array(args['annotation_class_ids'])
+        annotations_format = args.get('format', 'geojson')
+
+        # Compute the cartesian product of image_ids and annotation_class_ids
+        image_ids_grid, annotation_class_ids_grid = np.meshgrid(image_ids, annotation_class_ids, indexing='ij')
```
choosehappy, line 161: itertools is probably a bit more natural here? https://docs.python.org/3/library/itertools.html#itertools.product

quickannotator/api/v1/annotation/routes.py
```diff
@@ -129,4 +139,78 @@ def post(self, args):
...
+@bp.route('/export/server')
+class ExportAnnotationsToServer(MethodView):
+    @bp.arguments(server_models.DownloadAnnsArgsSchema, location='query')
+    def post(self, args):
+        """ Export annotations for multiple images and annotation classes to the server """
+
+        image_ids = args['image_ids']
+        annotation_class_ids = args['annotation_class_ids']
+        format = args.get('format', 'geojson')
+        # write_annotations_to_tarfile(image_ids, annotation_class_ids, format)
```
choosehappy, line 187: guessing this isn't done yet?

quickannotator/api/v1/annotation/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        self.progress += 1
+
+    def get_progress(self):
+        return (self.progress / self.total) * 100
+
+
+@ray.remote
+class AnnotationExporter:
+    def __init__(self, image_ids, annotation_class_ids, progress_actor):
+        image_ids_grid, annotation_class_ids_grid = np.meshgrid(image_ids, annotation_class_ids, indexing='ij')
```
choosehappy, line 32: same question regarding itertools

quickannotator/api/v1/annotation/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        for image_id, annotation_class_id in self.id_pairs:
+            with get_session() as db_session:
+                image_name = get_image_by_id(image_id).name
+                dsa_item = client.get_item_by_name(folder_id, image_name)
+                if dsa_item is None:
+                    raise Exception(f"Item with name {image_name} not found in folder {folder_id}")
+                
+                dsa_item_id = dsa_item['_id']
+                store = AnnotationStore(image_id, annotation_class_id, True, False)
+                feature_collection = store.get_all_annotations_as_feature_collection()    # TODO: this should probably be done with tempfile to avoid memory issues
```
choosehappy, line 53: indeed - this whole thing should be done either via tempfiles, or in a streaming manner, otherwise we get in trouble pretty quickly if there is a large data class for export :-\

quickannotator/db/crud/annotation.py
```diff
@@ -49,6 +62,37 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+        annotations (List[models.Annotation]): A list of annotation objects.
+
+    Returns:
+        geojson.FeatureCollection: A GeoJSON FeatureCollection containing the annotations.
+    """
+    features = []
+    for annotation in annotations:
+        feature = geojson.Feature(
+            id=annotation.id,
+            geometry=geojson.loads(annotation.polygon),
```
choosehappy, line 79: is this a place to use e.g., urjson or a faster library?

quickannotator/db/crud/annotation.py
```diff
@@ -49,6 +62,37 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+                # 'tile_id': annotation.tile_id,
+                # 'centroid': geojson.loads(annotation.centroid),
+                # 'area': annotation.area,
+                # 'custom_metrics': annotation.custom_metrics,
+                # 'datetime': annotation.datetime.isoformat()
+                'objectType': 'annotation'
+
+            }
+        )
+        features.append(feature)
```
choosehappy, line 91: whenever seeing a .append, this would suggest to use list comprehension?

quickannotator/db/crud/annotation.py
```diff
@@ -49,6 +62,37 @@ def get_annotation_query(model, scale_factor: float=1.0) -> Query:
...
+    
+    return geojson.FeatureCollection(features)
```
choosehappy, line 93: do you have an estimate of the compute time of this function? seems like it'll be intense if we have e.g., a million objects?

quickannotator/db/crud/annotation.py
```diff
@@ -122,22 +168,63 @@ def insert_annotations(self, polygons: List[BaseGeometry], tile_ids: List[int] |
...
+    
+
+    def export_all_annotations_to_tar(self, tarpath: str):
+        """
+        Saves all annotations to a tar archive.
+
+        Args:
+            tarname (str): The name of the tar file to save the annotations to.
+        """
+        with tarfile.open(tarpath, mode='w') as tar:
```
choosehappy, line 204: wrap in a gzip

choosehappy: you can see some examples here: https://andrewjanowczyk.com/serialization-and-storage-of-geojson-for-in-digital-pathology/

choosehappy: actually, you may be able to just use:

>>> fp = tarfile.open("zen.tar.gz","w:gz")

quickannotator/db/crud/annotation.py
```diff
@@ -122,22 +168,63 @@ def insert_annotations(self, polygons: List[BaseGeometry], tile_ids: List[int] |
...
+            annotations = self.get_all_annotations()
+            feature_collection = anns_to_feature_collection(annotations)
+            feature_collection_json = geojson.dumps(feature_collection)
+
+            # Create a tarinfo object for the GeoJSON file
+            tarinfo = tarfile.TarInfo(name=f"{self.get_annotation_table_name()}.geojson")
+            tarinfo.size = len(feature_collection_json)
```
choosehappy, line 211: i don't think this is correct, because the actual len/size will depend on the coding used, so this is more correct:

len(feature_collection_json.encode('utf-8'))

however, if you do that, you'll be doing it twice, here and again on line 214, so this should be done once and then reused in both places

quickannotator/db/utils.py
```diff
@@ -13,4 +16,96 @@ class DynamicAnnotation(base):
...
+        return os.path.join(self.nas_read, "images")
+
+    def get_input_masks_dir(self):
+        """
+        Get the directory path for annotation masks.
+
+        Returns:
+            str: The directory path containing annotation masks.
+        """
+        return os.path.join(self.nas_read, "masks")
```
choosehappy, line 52: very nice

quickannotator/db/utils.py
```diff
@@ -13,4 +16,96 @@ class DynamicAnnotation(base):
...
+        Get the directory path for project images.
+
+        Args:
+            proj_id (int): The project ID.
+            img_id (int): The image ID.
+
+        Returns:
+            str: The directory path for the project images.
+        """
+        return os.path.join(self.nas_write, "projects", f"proj_{proj_id}", "images", f"img_{img_id}")
```
choosehappy, line 65: exactly - best practice

choosehappy (changes requested):
some comments in here

quickannotator/db/crud/annotation.py
jacksonjacobs1: done.

quickannotator/db/crud/annotation.py
jacksonjacobs1: done.

quickannotator/db/crud/annotation.py
jacksonjacobs1: I have a feeling our best option is to progressively construct the feature collection using ogr2ogr like so (from chatgpt):

image

This will cap the memory footprint of the operation and allow us to avoid `FeatureCollection()`. I'm not sure of the compute time, but it's most likely O(N) .

Thoughts on the above approach? ChatGPT noted that with the python wrapper for ogr, you have to be careful not to call certain functions as some load the entire layer into memory with no feedback.

quickannotator/db/crud/annotation.py
jacksonjacobs1: Potentially - I'll leave a TODO here but it might be premature optimization - 
1. We don't know if this operation is the main performance bottleneck
2. Looking at your [blog post](https://andrewjanowczyk.com/serialization-and-storage-of-geojson-for-in-digital-pathology/), there's a ~50% read improvement the native json read method. This is substantial, but not an order of magnitude better

The uncompressed write operation is MUCH faster with orjson than native json. However, we generally don't serialize geojson anywhere (ST_AsGeoJSON returns serialized json by default).

Thoughts?

quickannotator/db/crud/annotation.py
choosehappy: can we agree to write a gz version of the geosjon? its likely at minimum a 70% reduction in size, so pretty worth it. honestly - this kind of thing makes me feel uncomfortable. i dislike "rolling my own" json dynamically - since any failure/edge-case can potentially result in a e.g., 3gb file that is entirely unparseable. the thing needs to be *tight*, one has to consider the case where some user does something which has { / [ , or whatever in some of the fields - are they properly escaped? that sort of stuff. it is...error prone.

that said, i don't know that i have a better solution in this case. an open question would be if this for "feature in layer " is a generator or if it pulls in the entire table into memoroy (as your comment hints at), which if it does, defeats a lot of the optimization potential. 

seems like if we wanna go big, our hand is kind of forced here?

quickannotator/db/crud/annotation.py
```diff
@@ -122,22 +162,68 @@ def insert_annotations(self, polygons: List[BaseGeometry], tile_ids: List[int] |
...
+        Args:
+            tarpath (str): The path of the tar file to save the annotations to.
+        """
+        # Ensure the directory for the tar file exists
+        os.makedirs(os.path.dirname(tarpath), exist_ok=True)
+
+        mode = 'w:gz' if tarpath.endswith('.gz') else 'w'
+
+        with tarfile.open(tarpath, mode=mode) as tar:
+            annotations = self.get_all_annotations()
```
choosehappy, line 204: so this code here w ould be replaced with what we discussed above?

quickannotator/dsa_sdk.py
```diff
@@ -0,0 +1,151 @@
...
+
+        Args:
+            base_url (str): The base URL of the DSA server.
+            api_key (str, optional): The API key to generate the Girder authentication token.
+            duration (int, optional): The duration (in days) for which the token is valid.
+        """
+        self.base_url = base_url.rstrip('/')
+        self.token = None
+        try:
+            self.get_token_with_api_key(api_key, duration)
```
choosehappy, line 19: should this set the token if successful?

choosehappy (commented):
looking good - a few points for (rapid) discussion/convergence

quickannotator/dsa_sdk.py
jacksonjacobs1: Currently the `get_token_with_api_key()` method internally sets `self.token` after a successful get. 

But what you're suggesting makes more sense - the `get_token_with_api_key` method should get the token and do nothing else. I've updated the code accordingly.

quickannotator/db/crud/annotation.py
jacksonjacobs1: > can we agree to write a gz version of the geosjon?

I think best practice is to gzip the entire tar archive (yielding .tar.gz, which is what we currently do) rather than zipping each item in the tar archive (json.gz). Is this what you meant?

> honestly - this kind of thing makes me feel uncomfortable. i dislike "rolling my own" json dynamically - since any failure/edge-case can potentially result in a e.g., 3gb file that is entirely unparseable.

This is a valid concern. Frankly, json is not designed for streaming to keep memory usage low. Here are some alternatives suggested by chatgpt:


Format | Streaming | Compression | Read Speed | Tooling (QGIS, etc) | Best For
-- | -- | -- | -- | -- | --
GeoJSON | ⚠️ Manual | ❌ | ❌ | ✅ | Small files, web
GPKG | ✅ Native | ✅ | ✅ | ✅ | Production-grade storage
FlatGeobuf | ✅ Native | ✅ | ✅✅ | ✅ | Web + large-scale
Parquet | ✅ Via libs | ✅✅ | ✅✅ | ❌ (not maps) | Data science workflows

However, none of these are supported by qupath (or DSA for that matter).

If we're married to json, we might want to just bite the bullet and implement our own streaming. The predictable structure of our geojson feature collection will play to our advantage here.

quickannotator/api/v1/annotation/routes.py
```diff
@@ -129,4 +132,70 @@ def post(self, args):
...
+        props_format = args['props_format']
+
+        resp = [{
+            "filepath": fsmanager.nas_write.global_to_relative(build_tarpath(image_id, annotation_class_id, is_gt)),
+            } for image_id, annotation_class_id in list(product(image_ids, annotation_class_ids))]
+
+        exporter = AnnotationExporter.remote(image_ids, annotation_class_ids)
+        exporter.export_remotely.remote()
+
+        return resp, 200
```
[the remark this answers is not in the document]
jacksonjacobs1: Yep. I've updated the code to return the actor name. I also added a utility function for naming actors:

```python
def compute_actor_name(project_id: int, type: constants.NamedRayActorType) -> str:
    """
    Generate a unique name for the actor using project ID and current timestamp.
    
    Args:
        project_id (int): The ID of the project.
        type (constants.NamedRayActorType): The type of the actor.
    
    Returns:
        str: The generated actor name.
    """
    timestamp = time.strftime('%Y%m%d%H%M%S')
    return f"{project_id}_{type.value}_{timestamp}"
```
also improved the response validation of these export endpoints.

quickannotator/db/crud/annotation.py
choosehappy: hmmm thinking more about this - what is the value of having a tar at all instead of simply providing links for each of the individual geojson files? bundling them all together into a single tar file is convenient for download, but its basically a nightmare for everything else : ) e.g., if i have 2 classes, 1 of them is 20GB and the other is 1kb, i download a single tar file, which i then have to untar (and/or decompress) in order to get access to the single 1kb file. why not provide each file individually so that they're not tighly coupled?  the situation you proposed (where everything is tar gziped together) is also less than ideal, since one has to process the entire 20GB file to get access to the 1kb file (since gzip is a streaming compression). at least if the gzs are "tarred" together, one can access the file sepereately (direct read out of tar), and then extract only that. still, my preference would be for more individual files (especially if we have something like a "manifest file" for rapid download ), as opposed to one big chunker. this should also provide us with some better caching options (E./g., class 1 changes, but class 2 doesn't change, only need to rexport class 1 then -- the class 2 file url can remain the same). thoughts?


ultimately - wen eed to support qupath and DSA - i'd be super annoyed if i exported 20gb of stuff and then had to "convert" it somehow to another format. if we lose the qupath compatability, we'd use a different approach all togehter (e.g., msgpack), to get even better compression.

>Another interesting aside is that qupath and DSA are actually incapable of displaying annotation sets at this scale (1million). 

is this true of qupath? i've definitely loaded in ~700k nuclei annotations on my laptop. the loading takes a minute or two, but once loaded seems to work okay - did you have some issue scaling to 1m?

sounds like you wanna bite the bullet on this and manually construct --- i don't know that i have a better solution : ) so  probably go with that approach and we can come back later if we need to make other decisions?

quickannotator/client/src/helpers/config.ts
```diff
@@ -32,10 +32,20 @@ export const MODAL_DATA: { [key: string]: ModalData } = {
...
 
 export const POPOVER_DATA = {
 
-}
\ No newline at end of file
+}
+
+export const TISSUE_MASK_CLASS_ID = 1;
+
+export const SERVER_URL = 'http://localhost:5000'; // Your Flask backend
```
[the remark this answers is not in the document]
choosehappy: i'm a bit nervous about having this sort of "configuration" stored in another other file. in the sense that, we have a config file for the backend, and now there is a config file for the front end -- some of them are constants which is fine because they're not expected to change -but in this case i'm wondering is there ever a situation where the user is ever going to need to change this value? or is it *really* a constant constant?

quickannotator/db/crud/annotation.py
choosehappy: the only thing i dislike about the tar is that still folks would need double storage. like, if i wanted to import my annotations into qupath, not only do i have to download them as a tar, but then i have untar it, and THEN i can import it. if i end up downloading 10 images with 10 classes, i have 100 tar files i need to "untar" to use downstream. its not...elegant...for the downstream user. i'm wondering - since we'll give a manifest file anyway, why not just keep all the files seperate and name them better? like class1.geojson.gz, class1.metadata.gz or whatever, and let the users download what they want?

for qupath - it depends highly on the machine - the annotations are stored in a java object so can be quite memory intensive. unlike e.g., dsa which i think is a front-end technology limitation (the browser itself has some limits, regardless of compute power), qupath can "scale" if you just get a better machine.

jacksonjacobs1: Good point! 

My one question is, if we return compressed geojson to the user don't they need to uncompress anyways before they can import into qupath?

extracting files from a compressed tar archive is also a one-step process from the user's perspective, so I'm not sure I see where this improves the user experience.

choosehappy: in python, you can load a compressed geoson.gz file directly. in qupath this is also possible with groovy (in a blogpost of mine somewhere) - now i'm actually wondering if  it can drag/drop accept a gz file? 


yes....yes it does

image


100% confident now that geojson.gz is the right approach : )

jacksonjacobs1: Awesome! I'm happy to move forward with this approach then.

jacksonjacobs1:
QuickAnnotator now exports individual formats separately. Each file saved is gzipped. The AnnotationsExportModal has also been simplified for better export format selection.

image

quickannotator/api/v1/annotation/utils.py
choosehappy: was this done? it looks like this is still meshgrid instead of itertools?

quickannotator/api/v1/annotation/utils.py
choosehappy: so this part here still will generate a whole file in memory first before transmitting, is that correct? is there an opportunity to do this with something like a generator?

quickannotator/db/crud/annotation.py
choosehappy: does this code reflect the final version? i thought in the end you were going to build your own geojson file, object by object and write it out in a streaming format? i may be misunderstanding this code, but i think these functions are generating the complete list in memory before writing out?

choosehappy (changes requested):

quickannotator/db/crud/annotation.py
jacksonjacobs1: Sorry, still working on this - 

One note is that the python bindings for GDAL (ogr2ogr) depend on `libgdal`, a library that would add 500MB to the docker container. 
https://pypi.org/project/GDAL/

Is this okay with you or should I look for a different solution for streaming polygons to a geojson file?

choosehappy: ouch - thats a lot of bloat for a single function....and not really a core function at that - i guess one question is how big is the current docker container (approxrimately), if its a 1% difference, then do it, if its a 50% difference then we should probably invest at least a few minutes looking for leaner alternatives?

choosehappy: hmmm not very interested in using subprocess - too flakey and hard to control (with poor error reporting / logging). especially hard in client/server paradigms - if those are the only options, happy to pay the extra 400mb for stability

jacksonjacobs1: There's fiona, which pip installs in 16MB

"Fiona depends on [GDAL](https://gdal.org/) but is different from GDAL’s own [bindings](https://gdal.org/api/python_bindings.html). Fiona is designed to be highly productive and to make it easy to write code which is easy to read."

https://pypi.org/project/fiona/

Interestingly fiona seems to handle geojson schema internally, while also streaming to keep memory low. Here's the snippet from chatgpt. Will have to confirm this of course
```python
import tempfile
import fiona
from fiona.crs import from_epsg

def export_to_geojson_using_fiona(pg_conn_str, table_name):
    # PostGIS connection string in the format Fiona expects:
    # "PG:host=... dbname=... user=... password=... port=..."
    source_uri = f"PG:{pg_conn_str}"

    # Open the PostGIS layer with Fiona
    with fiona.open(source_uri, layer=table_name) as source:
        # Create a temporary GeoJSON file
        with tempfile.NamedTemporaryFile(delete=False, suffix=".geojson", mode='w') as tmpfile:
            # Define schema and CRS for the output GeoJSON
            schema = source.schema.copy()
            crs = source.crs

            # Open a new Fiona collection to write GeoJSON features
            with fiona.open(
                tmpfile.name, 
                'w',
                driver='GeoJSON',
                schema=schema,
                crs=crs
            ) as dest:
                for feature in source:
                    dest.write(feature)

            print(f"GeoJSON written to: {tmpfile.name}")
            return tmpfile.name

```

jacksonjacobs1:
That's true. GDAL python looks rock solid. I have no problem moving forward with that library.

One note is that none of these libraries interact with sqlalchemy, though they do support bost pgis and spatialite directly.

choosehappy:
shhhhhhhhhake it, and then bbbbbake it

jacksonjacobs1:
@choosehappy everything should be done now, please take a look.

Note that we are actually able to execute the custom query on the ogr DataSource without loading polygons into memory. 

This is the safest solution for applying polygon scaling. Note that the custom query also returns the GEOM fields As_GeoJson, so there was actually no need to use the GIS functionality of ogr. We just call feature.getField("polygon"), load this into json, and then dump the geojson feature into the gzipped export file.

Alternatively I could have defined a second custom query without ST_AsGeoJson, and used ogr's `geom.ExportToJson()` method, but this seemed like a slightly more complicated way to accomplish the same thing.

quickannotator/api/v1/annotation/utils.py
```diff
@@ -0,0 +1,117 @@
...
+class ProgressTracker:
+    def __init__(self, total: int):
+        self.total = total
+        self.progress = 0
+        self.lock = threading.Lock()  # Add a threading lock for thread safety
+
+    def increment(self):
+        with self.lock:  # Ensure thread safety by locking during increment
+            self.progress += 1
+
```
choosehappy, line 28: technically speaking with this design its possible to get a progress > total, is that something to be concerned about? my intuition is no and its the "clients" job to keep track of their stuff and manage the state "correctly"?

quickannotator/api/v1/annotation/utils.py
```diff
@@ -0,0 +1,117 @@
...
+    def get_progress(self) -> float:
+        with self.lock:  # Lock when accessing progress
+            return (self.progress / self.total) * 100
+
+
+@ray.remote(max_concurrency=2)  # Add max_concurrency=2
+class AnnotationExporter(ProgressTracker):  # Inherit from ProgressTracker
+    def __init__(self, image_ids, annotation_class_ids):
+        self.id_pairs = [(int(image_id), int(annotation_class_id)) for image_id, annotation_class_id in product(image_ids, annotation_class_ids)]
```
choosehappy, line 37: nnnnnnnnnnnnice

quickannotator/api/v1/annotation/utils.py
```diff
@@ -0,0 +1,117 @@
...
+                    raise Exception(f"Item with name {image_name} not found in folder {folder_id}")
+                
+                dsa_item_id = dsa_item['_id']
+                store = AnnotationStore(image_id, annotation_class_id, True, False)
+                feature_collection = store.get_all_annotations_as_feature_collection()    # TODO: this should probably be done with tempfile to avoid memory issues
+                feature_collection_json = geojson.dumps(feature_collection)
+                fc_bytes = feature_collection_json.encode('utf-8')
+                upload_id = client.post_file(
+                    parent_id=dsa_item_id,
+                    file_id=hex(int(dsa_item_id, 16) + 1)[2:].zfill(24),  # TODO: this is janky, ideally should be done by
```
choosehappy, line 63: done by what?

choosehappy (commented):
looks very good! a few comments for clarification

jacksonjacobs1:
All really good feedback. Should be incorporated now

quickannotator/api/v1/annotation/utils.py
choosehappy: you see this one?

quickannotator/db/__init__.py
```diff
@@ -16,8 +18,12 @@
 Base = declarative_base()
 Base.query = db_session.query_property()
 
+class Dialects(enum.Enum):
+    SQLITE = "sqlite"
```
choosehappy, line 22: nice

quickannotator/db/__init__.py
```diff
@@ -47,3 +53,23 @@ def get_session():
     finally:
         db_session.remove()  # Cleanup session automatically
         #print("Session closed")
+
+@contextmanager
+def get_ogr_datasource():
+    dialect = db_session.bind.dialect.name
+    if dialect == Dialects.SQLITE.value:
+        ogr_conn_str = f"SQLite:{engine.url.database}"
+    elif dialect == Dialects.POSTGRESQL.value:
```
choosehappy, line 62: beautiful

quickannotator/db/crud/annotation.py
```diff
@@ -122,22 +172,112 @@ def insert_annotations(self, polygons: List[BaseGeometry], tile_ids: List[int] |
...
+
+        annotations = self.get_all_annotations()
+        feature_collection = anns_to_feature_collection(annotations)
+        feature_collection_json_bytes = geojson.dumps(feature_collection).encode('utf-8')
+
+        # Save the GeoJSON data directly to a .geojson.gz file
+        with gzip.open(filepath, 'wb') as gz_file:
+            gz_file.write(feature_collection_json_bytes)
+
+    def export_to_geojson_file(self, filepath: str = None, compress: bool = False) -> str:
```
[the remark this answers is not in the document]
choosehappy: did you see this one?

choosehappy (approved):
bbbbbbbbbbbbeautiful

PR #107, opened 2025-07-10
[#97] Feature: only train on "complete" tile annotations

jacksonjacobs1:


jacksonjacobs1:
## Feature description:
QuickAnnotator currently greedily trains on ground truth annotations - as soon as a single annotation has been committed to a tile, the tile is considered valid for training.

However, in practice many tiles have more than one e.g., tubule, requiring multiple annotations.

Tiles should not be made available for training until the user is "done" annotating the tile. This event must be triggered by some kind of user input.


1. Tile is hidden from the training set whenever an annotation is modified in it.
1. The new/modified annotation is saved, and the new tile's state is updated, indicating that it has been updated but will not be used for training.
    - gt_counter = -1
    - gt_datetime is updated
1. In the UI, a border is drawn around the tile which has been hidden from the training set.
1. If the new annotation's tile_id is different from the old annotation's tile_id, the old tile is "unhidden" from the training set.
- Update the old tile's gt_counter to 0
- gt_datetime is not updated.

jacksonjacobs1:
We elected not to implement this feature. Rather, the DL component should be designed to account for noisy ground truth labels.

PR #102, opened 2025-07-08
Dl robustness

jacksonjacobs1:
# PR: Enhancements and Refactors for TileStore, Logging, Error Handling, and DB Compatibility

---

## 🧠 Deep Learning & TileStore Refactor
- Moved DL methods interacting with tiles into the `TileStore` abstraction.
- Made `TileStore` aware of database dialects (e.g., Postgres, SQLite).
- Conditionally applied mask intersection logic for tile ID retrieval based on annotation class.
- Enhanced fault tolerance with cache error handling.
- Built `try-catch` directly into cache manager.

---

## 🧹 General Refactoring & Cleanup
- Updated `ProjectPage` layout to use transparent backgrounds for list items.
- Fixed bug in `ViewportMap` tile ID check (explicit null comparison).
- Introduced `tileIdIsValid` utility for safer tile ID validation.
- Commented out dangerous file system tests.
- Removed stray breakpoint in `TileDataset`.

---

## 📦 File Path and I/O Improvements
- Introduced utilities for managing checkpoint and log file paths.
- Ensured directories are created if they do not exist.
- Refactored raw output saving logic and directory management.
- Removed `secure_filename` to preserve original uploaded filenames.

---

## 📝 Logging System Improvements
- Introduced `LoggingManager` and `LoggerNames` enum for consistent logging setup.
- Replaced print statements with logger calls in training, inference, and Ray components.
- Improved logging in annotation processes and core modules.

---

## 🌍 Frontend Enhancements
- Improved error handling and initialization in `ViewportMap`.
- Prevented rendering of `MASK_CLASS_ID` in UI components.
- Added a very basic confirmation alert when the user switches from the tissue mask class to another annotation class.
- Disabled toolbar buttons/radios to improve user experience.

---

## 🚫 Error Handling & Robustness
- Added error response for invalid annotation posts.
- Improved schema for annotations outside tissue masks.
- Added 404 handling for image deletion.

---

## 🧪 Testing
- Added robust tests for REST utilities (delete project/image/annotation class).
- Confirmed tests pass after annotation validation changes.

---

## 🛠️ Backend Enhancements
- Added `MASK_CLASS_ID` checks to prevent unwanted predictions.
- Improved token handling in `DSAClient`.
- Refactored cache mechanism and annotation query logic.
- Added `annotation_class_id` to image cache.

---

## 🗃️ Database Compatibility
- Made data loading logic SQLite/Postgres agnostic.
- Fixed `osgeo.ogr` connection string; PostGIS export now works.
[...]

PR #111, opened 2025-07-18
[#102] FEATURE: add brush & eraser tools

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/102

### DIFF
--- a/quickannotator/client/src/components/viewportMap.tsx
+++ b/quickannotator/client/src/components/viewportMap.tsx
@@ -6,7 +6,7 @@ import { Point, Polygon, Feature, Position, GeoJsonGeometryTypes } from "geojson
 
 import { TOOLBAR_KEYS, INTERACTION_MODE, LAYER_KEYS, TILE_STATUS, MODAL_DATA, RENDER_PREDICTIONS_INTERVAL, RENDER_DELAY, MAP_TRANSLATION_DELAY, MASK_CLASS_ID, COOKIE_NAMES, POLYGON_OPERATIONS, POLYGON_CREATE_STYLE, POLYGON_CREATE_STYLE_SECONDARY, IMPORT_CREATE_STYLE, BRUSH_CREATE_STYLE, BRUSH_CREATE_STYLE_SECONDARY, BRUSH_SIZE } from "../helpers/config.ts";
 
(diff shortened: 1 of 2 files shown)


quickannotator/api/v1/annotation/utils.py
```diff
@@ -123,4 +126,67 @@ def compute_actor_name(project_id: int, type: constants.NamedRayActorType) -> st
...
+        Raises:
+            ValueError: If either of the input polygons is not valid.
+        """
+        if not poly1.is_valid:
+            raise ValueError("First polygon is not valid")
+        if not poly2.is_valid:
+            raise ValueError("Second polygon is not valid")
+        result = poly1.union(poly2)
+
+        return result
```
choosehappy, line 158: is the main reason for this function to add consistent error checking?

quickannotator/api/v1/annotation/utils.py
```diff
@@ -123,4 +126,67 @@ def compute_actor_name(project_id: int, type: constants.NamedRayActorType) -> st
...
+        if result.is_empty:
+            return None
+        if not result.is_valid:
+            raise ValueError("Resulting polygon is not valid after difference operation")
+        if result.geom_type == 'MultiPolygon':
+            # If the result is a MultiPolygon, return the largest polygon by area
+            if multipoly_func == constants.MultiPolygonToPolygonFuncs.MAX:
+                result = max(result.geoms, key=lambda geom: geom.area)
+            else:
+                raise ValueError(f"Unsupported multipolygon function: {multipoly_func}")
```
choosehappy, line 190: this seems to be unnecessarily limiting, why not allow the string to define the function directly, e.g., 
```python
import builtins

try:
    func = getattr(builtins, multipoly_func)
except AttributeError:
    raise ValueError(f"Unsupported multipolygon function: {multipoly_func}")

result = func(result.geoms, key=lambda g: g.area)
```

so that the user can supply any string and it would run?

quickannotator/constants.py
```diff
@@ -20,6 +20,9 @@ class PolygonOperations(enum.IntEnum):
     UNION = 0
     DIFFERENCE = 1
 
+class MultiPolygonToPolygonFuncs(enum.IntEnum):
+    MAX = 0
```
choosehappy, line 24: see above - seems like it can be more general?

choosehappy (changes requested):
this Max seems like it could be made more general?

PR #122, opened 2025-11-19
Feature/101-feature-change-gray-flashing-predictions-generation

### LINKED ISSUE #83: [FEATURE] Improve client annotation rendering
## Description  
The current rendering approach does not perform equally across different zoom levels because:
1. We render a greater number of WebGL features while zoomed out
2. We render a greater number of points while zoomed out

A better rendering approach involves the following changes:
1. Decouple each WebGL feature from a single tile (and tile id). Each WebGL feature should be capable of including annotations from multiple tiles within a group.
2. GET Tile endpoints should accept a `grouping_factor` parameter, which causes a group_ids array to be returned.
3. Create a new endpoint for getting annotations within multiple tiles. This endpoint should also accept a simplification threshold so that resulting polygons are simplified to reduce the number of points to render.

e.g.,
Given a 4x4 tilegrid of ids 0-15, say we get all tiles, with grouping_factor set to 1 (defualt)

```
resp.tile_ids = [0 - 15]
resp.group_ids = [0 - 15]
```

With grouping_factor set to 2:

```
resp.tile_ids = [0 - 15]
resp.group_ids = 
[0,0,1,1,
0,0,1,1,
2,2,3,3,
2,2,3,3]
```

With grouping_factor set to 3:
```
resp.tile_ids = [0 - 15]
resp.group_ids = 
[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]
```

The client then only requests annotations for each unique group_id. Each group_id thus corresponds to a webgl feature.

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/101


choosehappy (approved):
looks good!

choosehappy (approved):

jacksonjacobs1:
image

Finished this PR. Changes noted below:
### Adaptive zoom
- Shapes and render batches adapt as the user zooms in and out. 

### Tile-level DL status improved
- Tile statuses are now shown in a separate layer. They are refreshed every 8 seconds alonside the predicted annotations, determined by `RENDER_PREDICTIONS_INTERVAL`, a configurable variable.
- Gray: tiles in the `STARTPROCESSING` state (e.g., pending DL processing)
- Yellow: tiles in the `PROCESSING` state (i.e., claimed by a DL actor for processing)
- Green: tiles in the `DONEPROCESSING` state (i.e., tiles are processed)

> Note: predictions are now shown regardless of what state the tile is in. If the tile status is in one of the pending states (e.g, `STARTPROCESSING` or `PROCESSING`), the displayed predictions are shown but may not reflect the latest DL model. If the the tilestatus is `DONEPROCESSING`, the predicted annotations are somewhat recent (at most 1 minute old, determined by `constants.TILE_PRED_EXPIRE`, a configurable variable.

### TODO
A second pull request will be created adding toggles for each of the three layers.

PR #124, opened 2025-11-24
Task/194-feature-add-sphinx-docs-to-quickannotator

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/194


jacksonjacobs1:
You can review the current state of the docs here:
https://quickannotator.readthedocs.io/en/latest/

choosehappy (approved):
looks good! i think my only question is - does the user need to run quickannotator/npm themselves? it seems like since thsoe are awlays taking place they could be taken care of by the docker container itself?

jacksonjacobs1:
Sure, this will be added in #122

PR #125, opened 2025-12-20
[#157] FEATURE: toggle train pred loop

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/157


quickannotator/dl/training.py
```diff
@@ -164,7 +164,7 @@ def train_pred_loop(config):
             run_inference(device, model, tiles)
             
         logger.info(f"No more STARTPROCESSING tiles for annotation class {annotation_class_id}. Entering training loop.")
-        if ray.get(myactor.getEnableTraining.remote()):
+        if ray.get(myactor.get_enable_training.remote()):
```
choosehappy, line 173: we just need to make sure that this isn't a very slow call - i don't tihnk it is and i remember testing it back in the day - but this is doing a lot of work behind the scenes  (remote calls, etc) - so if it takes a long time, the entire trianing loop will be much less effective. its something we should keep an eye on,

choosehappy (approved):

PR #126, opened 2026-01-05
Feature/159 feature gt pane filter visibility buttons

jacksonjacobs1:
[#159] FEATURE: gt pane filter/ visibility buttons

https://jacksonjjacobs.com/openproject/work_packages/159


jacksonjacobs1:
image

## Features added:
### 1. Layer toggles
Layer toggles hide layers rather than clearing them, allowing for instantaneous toggling rather than having to hit the backend for annotations.

image


### 2. Styled tooltips:
image

### 3. True parallel, progressive rendering of tiles
Asyncronous functions like REST requests are wrapped in async guards to protect against race conditions.
> NOTE: an optional improvement over this could be an abortController which passes an abort signal to the javascript's fetch API.

### 4. Display annotation count in GT and Pred annotation lists:
image

### 5. Collapse annotation lists alongside layers:
image

### 6. Conditional prediction requests
Only request predictions for tiles when either the predictions layer or the tile status layer is enabled.


## Potential additional features
1. A circular progress bar in the toolbar indicating how long until the next automatic re-render of predictions. Next to the progress bar, a dropdown could allow the user to select how often they would like predictions to refresh

choosehappy (approved):
a bit out of my depth with all the tsx : ) looks.....i mean..does it work? ha

choosehappy:
These screenshots look amazing!! Great job!

PR #129, opened 2026-02-10
V2.0 dlfeat: tweaks

jacksonjacobs1:
Made the following changes to the DL code:
1. Substituted `_hierarchical_prototype_loss` function from the prototpying code into loss.py
2. Keep loaded images at fp32 precision for training.
3. Updated mask cache: weight object no longer relevant.
4. Off-screen tiles are set back to the UNSEEN state whenever the viewport bbox changes, forcing DL actors to process on-screen tiles.

### DIFF
--- a/quickannotator/dl/loss.py
+++ b/quickannotator/dl/loss.py
@@ -57,31 +57,6 @@ def edge_loss(positive_mask: torch.Tensor, pred: torch.Tensor) -> torch.Tensor:
     return bce_edge
 
 
-def masked_mse_loss(preds: torch.Tensor, targets: torch.Tensor, mask: torch.Tensor) -> torch.Tensor:
-    """
-    Compute masked MSE loss.
-
-    Args:
-        preds: Predictions.
-        targets: Targets.
-        mask: Mask to apply.
-
-    Returns:
-        Masked MSE loss.
-    """
-    criterion = nn.MSELoss(reduction='none')
-    per_element_loss = criterion(preds, targets)
-
-    # Expand mask to match channel dimension if needed
-    if mask.dim() < per_element_loss.dim():
-        mask = mask.unsqueeze(1)  # [B,1,H,W]
-
-    # Apply mask
-    masked_loss = per_element_loss * mask
-
-    # Average over valid pixels, avoid division by zero
-    return masked_loss.sum() / mask.sum().clamp(min=1)
-
 
 def total_variation_loss(mask: torch.Tensor) -> torch.Tensor:
     """
@@ -93,8 +68,8 @@ def total_variation_loss(mask: torch.Tensor) -> torch.Tensor:
     Returns:
         Total variation loss.
     """
-    dx = torch.abs(mask[:, :, 1:, :] - mask[:, :, :-1, :])
-    dy = torch.abs(mask[:, :, :, 1:] - mask[:, :, :, :-1])
+    dx = torch.abs(mask[:, :, 1:, :] - mask[:, :, :-1, :]) ** 2
+    dy = torch.abs(mask[:, :, :, 1:] - mask[:, :, :, :-1]) ** 2
     return (dx.mean() + dy.mean())
 
 


quickannotator/api/v1/tile/routes.py
```diff
@@ -39,12 +39,17 @@ class PredictTiles(MethodView):
     def post(self, args, image_id, annotation_class_id):
         """     stage a tile for DL processing     """
         tilestore: TileStore = TileStoreFactory.get_tilestore()
+
+        # Reset all STARTPROCESSING tiles to UNSEEN before staging new tiles. Note that this results in 2 transactions to the Tile table.
+        tilestore.reset_pending_tiles(annotation_class_id)
```
choosehappy: its not obvious to me that this should be bundled in here or have its own route to be an explicit option? this would suggest that if i submit 2 predict tiles posts, the first one will automatically be cancelled - which removes the notion of e.g., stateless APIs. at minimum i think there should be another parameter, defaults to false, which resets current pending. hwoever, this this does require a roundtrip for db access anyway - perhaps it justiifes another end point?

quickannotator/dl/training.py
```diff
@@ -181,31 +181,25 @@ def train_pred_loop(config):
...
             
             # Unpack patch batch: (patch_image, patch_mask, hv_map)
-            images = batch_data[0]
-            masks = batch_data[1]
-            hv_maps = batch_data[2]
+            images, masks, hv_maps = batch_data
             
             # Move to device and normalize images to [0, 1]
-            images = images.half().to(device) / 255.0
+            images = images.to(device) / 255.0
```
choosehappy: this makes me a bit sad - is this really the cause of some instaibility / NaNs? the throughput in transferring is worth it - 2x as fast!

quickannotator/dl/loss.py
```diff
@@ -625,6 +628,7 @@ def forward(
         
         # Safety check for NaN/Inf losses
         if torch.isnan(loss_total) or torch.isinf(loss_total):
+            logger.error(f"NaN/Inf detected in total loss computation: {losses}")
```
choosehappy: this should likely be expanded -- if its nan then set to zer0 -- don't let it propagate into the system. if all of the losses have such a condition independently supplied, its likely that that will prevent non-sane situations to begin with - since no nan could enter the system, and after update of the weights, the liklihood of another nan generation is small?

quickannotator/dl/loss.py
```diff
@@ -634,60 +638,67 @@ def _hierarchical_prototype_loss(
...
-        
-        # Normalize embeddings
-        obj_emb_norm = F.normalize(obj_embeddings, dim=1)  # (B, D, H, W)
-        
-        loss_total = 0.0
-        valid_samples = 0
-        
+
+        # Resize mask ONCE
+        mask = F.interpolate(
```
choosehappy: hmm...i can't seem to find where the normalization of the embeddings takes place?

quickannotator/dl/loss.py
```diff
@@ -634,60 +638,67 @@ def _hierarchical_prototype_loss(
...
-        
-        return loss_total / max(valid_samples, 1)
+
+            emb_b = emb[b][pos]  # (N_b, D)
+            proto_b = emb_b.mean(dim=0, keepdim=True)
+            img_loss = img_loss + ((emb_b - proto_b) ** 2).sum(dim=1).mean()
+            valid_imgs += 1
+
+        if valid_imgs > 0:
+            img_loss = img_loss / valid_imgs
```
choosehappy: although admitadely i've only been looking at this for a few minutes - i don't see how the previous and current version are mathematically equivalet in their result? e.g., where is the infonce style loss for negative pixels ?

choosehappy (changes requested):
I have some concerns here - is the pre/post commit versions really mathematically equivalent? it looks like .e.g, the prototype loss was removed, in particular for negative pixels?

quickannotator/dl/loss.py
jacksonjacobs1: That's a good point. They look quite different to me too.

How was the original _hierarchical_prototype_loss() function in loss.py generated?  The original code was producing errors that did not seem trivial to resolve. The code that I substituted in was from the `hierarchical_prototype_loss()` function from `quickannotator/dl/research_work/improved-v1/losses.py` which I believe is the code that you validated. 

If these two functions are logically inequivalent, and one was the LLM-generated successor of the other, that would be a bit concerning.

quickannotator/api/v1/tile/routes.py
jacksonjacobs1: If these are separated into two endpoints, the new "reset" endpoint would need to accept "tile_ids" as a parameter so that the specified tile ids are (or are not) reset. This is because we can't assume any ordered relationship between REST requests. The "reset" request could be fulfilled after the "predict" request, nullifying the effect of "predict" unless "reset" ignores the tile ids that "predict" operates on.

quickannotator/dl/loss.py
jacksonjacobs1: Okay. Are we confident that the likelihood of another nan generation is small after the first one? Or does NaN output -> set total loss to 0 -> update weights just lead to more NaN output?

I thought in a previous comment you said once total loss is NaN, the model is already non-sane?

choosehappy: I am unsure we're talking about the same thing --- here is an example. I can put a Nan _forward_ through a network and it remains sane:


image



because forward passes are essentially somehow "read only"

however, if i do a backpass which has a Nan anywhere in it - the Nan propogates like a virus and makes the model not sane anymore:

image


There should hopefully never be an input image that has a Nan, otherwise something has gone catastrophically wrong with reading a WSI -- but perhaps you want to add a check for that

however, the goal here is to, 100% of the time, prevent any loss from being involved in a back propogation which has yielded a Nan. by setting it to 0 - it essentially turns it off, so that its contribution doesn't propogate the Nan, and the network remains "sane". the system will still take a meaningful step, which hopefully won't cause whatever error again -- but even if it does, it would still be caught and set to zero. 

do you see what i mean now?

jacksonjacobs1: Ah okay, makes sense now.

I'll "expand" the NaN check to each individual loss and set the individual loss to zero if it becomes NaN.

quickannotator/api/v1/tile/routes.py
choosehappy: > This is because we can't assume any ordered relationship between REST requests.

i would have guess that the UI is supposed to be orchestrating this sort of a thing? it should know which one it requests for predict, and which ones are out of the viewpane?

> he new "reset" endpoint would need to accept "tile_ids" as a parameter so that the specified tile ids are (or are not) reset. 

I don't see where tile_ids are used here in the actual reset functionality? it looks to reset everything by annotation_class_id

https://github.com/choosehappy/QuickAnnotator/pull/129/changes/BASE..70d5cdd891c0959fee76f3b081a83fbddbc82398#diff-c4ddf7bf2d0af4f3a7a70f6f7af51a94430ccd15957c415dd4ac1e1dbf4da764R274-R280



Overall i'm not entirely sure how this works - it seems like if the user is panning around, and tiles are getting requested for prediction, but then they tile next to it, we stop the prediction (reset all), then add one new tiles, but then reset them all --- i guess i'm just unclear when the prediction will actually be allowed to fully execute? is it only in the case when the user doesn't move around? then again, i'm not looking at the front end code so perhaps there is a mechanism there coordinating all this?

quickannotator/dl/loss.py
choosehappy: my original code is here, in in quickannotator/dl/loss.py

https://github.com/choosehappy/QuickAnnotator/pull/127/changes#diff-78c84d9686dc8d9efb052b17df57a22bdbeb7c2896fccb3250ef097ca3869f51R674


note however there are two places for sim_pos 

one in WeaklySupervisedSegmentationLoss

one in HierarchicalPixelContrastiveLoss

where each does a different thing - i see two possibiltiies

1) github didn't align the files correctly
2) if/when you LLMed it, it conflated the two and yielded a frankenstein which is neither on?


did it try to merge both classes and get lost on what their intended functionality is?

quickannotator/api/v1/tile/routes.py
jacksonjacobs1: > I don't see where tile_ids are used here in the actual reset functionality? it looks to reset everything by annotation_class_id

That's right. We currently reset all tiles for the annotation class to UNSEEN before setting specific tiles to STARTPROCESSING.

> Overall i'm not entirely sure how this works - it seems like if the user is panning around, and tiles are getting requested for prediction, but then they tile next to it, we stop the prediction (reset all), then add one new tiles, but then reset them all --- i guess i'm just unclear when the prediction will actually be allowed to fully execute? is it only in the case when the user doesn't move around? then again, i'm not looking at the front end code so perhaps there is a mechanism there coordinating all this?

Just looked back and I think this code was a suboptimal solution to a problem I observed. The code works well, but is accounting for behavior that results from a bug.

Here's the problematic code:
```python
def upsert_pred_tiles(self, image_id: int, annotation_class_id: int, tile_ids: List[int], pred_status: TileStatus=TileStatus.UNSEEN, process_owns_tile=False, reset_pending=False) -> List[db_models.Tile]:
    """
    Inserts new prediction tiles or updates existing prediction tiles in the database.
    This function attempts to insert new prediction tiles with the given annotation_class_id, image_id, and tile_ids.
    If prediction tiles with the same annotation_class_id, image_id, and tile_ids already exist, it updates the relevant fields.

    Args:
        image_id (int): The ID of the image.
        annotation_class_id (int): The ID of the annotation class.
        tile_ids (List[int]): The IDs of the tiles.
        pred_status (TileStatus): The status of the tile prediction.
        process_owns_tile (bool, optional): If True, updates tiles regardless of their current status. Defaults to False. Only a ray actor should set this to True.

    Returns:
        List[models.Tile]: The tiles that were inserted or updated.
    """

    if pred_status is None:
        raise ValueError("pred_status cannot be None.")

    # Reset all STARTPROCESSING tiles to UNSEEN. Note this is a separate transaction which adds some exposure to benign race conditions.
    if reset_pending:
        self.reset_pending_tiles(annotation_class_id)

    current_time = datetime.now()

    insert_fields = {
        'pred_status': pred_status,
        'pred_datetime': current_time
    }
    
    if process_owns_tile:
        update_fields = insert_fields
    else:
        expiration_thresh = current_time - timedelta(minutes=constants.TILE_PRED_EXPIRE)

        tile_is_unseen_or_stale = case((db_models.Tile.pred_status == TileStatus.UNSEEN, 1),
            (
                (db_models.Tile.pred_status == TileStatus.DONEPROCESSING) & 
                (db_models.Tile.pred_datetime <= expiration_thresh), 1
            ),
            else_=0
        )

        update_fields = {
            'pred_status': case((tile_is_unseen_or_stale == 1, pred_status), else_=db_models.Tile.pred_status),
            'pred_datetime': case((tile_is_unseen_or_stale == 1, current_time), else_=db_models.Tile.pred_datetime)
        }

    tiles = self._upsert_tiles(
        image_id=image_id,
        annotation_class_id=annotation_class_id,
        tile_ids=tile_ids,
        insert_fields=insert_fields,
        update_fields=update_fields
    )
    return tiles
```

The problem is that we don't update pred_datetime for tiles already in the STARTPROCESSING state. With this bug, when the user moves the viewport to tiles that were already in the STARTPROCESSING state, these tiles never get updated datetimes and get stuck at the back of the queue of tiles to process.

I'll fix this bug and will removed the "resetting" logic added in this PR.

quickannotator/dl/loss.py
jacksonjacobs1: I used an LLM very lightly (might have even just copy pasted) to replace the code in _hierarchical_prototype_loss() with the code from here
https://github.com/jacksonjacobs1/QuickAnnotator/blob/70d5cdd891c0959fee76f3b081a83fbddbc82398/quickannotator/dl/research_work/improved-v1/losses.py#L86

Can you clarify what the code inside`v1_improved` is? I was under the impression that you used this code to produce the results on Fan's data, whereas the modifications to loss.py, training.py etc. were generated by an LLM and were not tested.

quickannotator/dl/training.py
jacksonjacobs1: Anecdotally, NaN errors were being raised much less infrequently after removing this .half() call.

choosehappy: oh yes, i very much believe you - the half precision and amp can induce instability in poorly designed systems  - but normally when its happened to me in the past (getting Nans) - it was due to incorrect system settings, e.g., LR was too high, or some other error in a loss function -- i can't really remember anytime it was definitively due just to "rounding errror" - i.e., it has always been fixable - which is why we use it extensively in hoverfast:

https://github.com/search?q=repo%3Achoosehappy%2FHoverFast%20half&type=code

quickannotator/dl/loss.py
choosehappy: yes, that is what v1_improved is - but i think the wrong snippet was copy and pasted?

the code which github is showing has been deleted is more similar to this:

https://github.com/jacksonjacobs1/QuickAnnotator/blob/70d5cdd891c0959fee76f3b081a83fbddbc82398/quickannotator/dl/research_work/improved-v1/losses.py#L368

noting that these are the same lines which were removed in your version:

https://github.com/jacksonjacobs1/QuickAnnotator/blob/70d5cdd891c0959fee76f3b081a83fbddbc82398/quickannotator/dl/research_work/improved-v1/losses.py#L442


image


this is to say: hierarchical_prototype_loss (a loss based on emebeddings, low dimensional embedding of features from the UNET ) != HierarchicalPixelContrastiveLoss (a loss based on pixels, higher dimensional)

This code here on the right hand side is definitely from a HierarchicalPixelContrastiveLoss  - is it possible this may be simply due to both having the word hierarchical in their name, many parts of their code do look very very similar, and thus were merged together?

jacksonjacobs1: The code on the right hand side (i.e., the code I pasted into loss.py) is from this function in losses.py.

Here's a permalink to the code from your latest commit:
https://github.com/choosehappy/QuickAnnotator/blob/e3dd5834081af379d718bec6820936495c6e991f/quickannotator/dl/research_work/improved-v1/losses.py#L86

In improved-v1/train.py, this function is used to compute the object-level embedding loss:
https://github.com/choosehappy/QuickAnnotator/blob/e3dd5834081af379d718bec6820936495c6e991f/quickannotator/dl/research_work/improved-v1/train.py#L296

Similarly, in loss.py the _hierarhical_prototype_loss() function is used to compute the object-level embedding loss. I took this to mean that these functions were interchangeable. 
https://github.com/choosehappy/QuickAnnotator/blob/e3dd5834081af379d718bec6820936495c6e991f/quickannotator/dl/loss.py#L592

Wait - are you sure the right-hand side of the diff is the HierarchicalPixelContrastiveLoss? Looking at losses.py, seems like the left hand side of the diff is from HierarchicalPixelContrastiveLoss
https://github.com/choosehappy/QuickAnnotator/blob/e3dd5834081af379d718bec6820936495c6e991f/quickannotator/dl/research_work/improved-v1/losses.py#L368

choosehappy: This is a disaster - my apologies!! it totally my fault!

its me that has made the error -- and when pushed, co-pilot admits that it was "its fault" - but naturally the error is mine because i reviewed it - but the code between all 3 things was so similar that i didn't see the difference at the time -- look at this conversation:

image

image

image



All that time saved....and now spent tracking down an error a human never would have made!

choosehappy: btw - i'm also wondering and suspecting that this may explain the Nans we saw before. as mentioned in my other comment - usually a coding bug and not an actual real-world instability..

quickannotator/dl/loss.py
choosehappy: Ignore - the normalization for this takes place in the model

quickannotator/dl/loss.py
choosehappy: This code looks exactly like what i had --- this was a lot of effort for me to say - looks good!  darn... again my apologies!

PR #130, opened 2026-02-17
V2.0 dlfeat

jacksonjacobs1:


deployment/docker-compose.yaml
```diff
@@ -31,21 +28,20 @@ services:
 
   qa:
     image: histotools/quickannotator:main
-    container_name: qa
     volumes:
-      - nas_read:/opt/QuickAnnotator/quickannotator/mounts/nas_read
+      - /home/jjaco34/digital_slide_archive/devops/dsa/assetstore/emory_ts_for_dsa:/opt/QuickAnnotator/quickannotator/mounts/nas_read
```
choosehappy: is this correct?

choosehappy (approved):
okay for me if you can confirm his one question

quickannotator/dl/dl_config.py
```diff
@@ -25,7 +25,7 @@ class DataConfig:
     max_patches_per_image: int = 200
     
     # Dataset
-    num_workers: int = 8
+    num_workers: int = 1
```
choosehappy: whats the motivation for making this 1? 1 is rarely the correct answer - its usually more efficiently to have either 0 (so everything is done in process and thus no communication overhead or management is needed) or a higher number which allows more parallelization?

quickannotator/dl/training.py
```diff
@@ -195,12 +196,17 @@ def train_pred_loop(config):
     # procRunningSince will be None if the DL processing is to be stopped, resulting in the model being unloaded from GPU and the training loop exiting
     while ray.get(myactor.get_proc_running_since.remote()):    
         tilestore = TileStoreFactory.get_tilestore()
-        while tiles := tilestore.get_pending_inference_tiles(annotation_class_id, batch_size_infer):
+        infer_iter = 0
+        while infer_iter <= max_inference_iterations:
```
choosehappy: not really in love with this - seems clunky. why not just accept a parameter on get_pending_inference_tiles with the # of desired tiles? i think this is a sql query, so one can simply put a "LIMIT" on it?

choosehappy: oh i see, this is based on the iterations not the # of tiles - i think the correct pattern is then a for loop:

```
for _ in range(max_inference_iterations + 1):
    tiles = tilestore.get_pending_inference_tiles(annotation_class_id, batch_size_infer)
    if not tiles:
        break
```

choosehappy (changes requested):
Some minor thoughts in here -please take a look

quickannotator/dl/dl_config.py
jacksonjacobs1: The issue is that the 8 concurrent workers are not only reading from the database, but writing to it as well. 
This causes the following error:
```
sqlalchemy.exc.DatabaseError: (psycopg2.DatabaseError) error with status PGRES_TUPLES_OK and no message from the libpq
[SQL: UPDATE tile SET gt_counter=CASE WHEN (tile.gt_counter < %(gt_counter_1)s) THEN tile.gt_counter + %(gt_counter_2)s ELSE tile.gt_counter END, gt_datetime=%(gt_datetime)s WHERE tile.id = (SELECT tile.id 
FROM tile 
WHERE tile.annotation_class_id = %(annotation_class_id_1)s AND tile.gt_datetime IS NOT NULL ORDER BY tile.gt_counter ASC, tile.gt_datetime ASC 
 LIMIT %(param_1)s FOR UPDATE SKIP LOCKED) RETURNING tile.id, tile.annotation_class_id, tile.image_id, tile.tile_id, tile.pred_status, tile.pred_datetime, tile.gt_counter, tile.gt_datetime]
[parameters: {'gt_counter_1': 5, 'gt_counter_2': 1, 'gt_datetime': datetime.datetime(2026, 4, 1, 11, 17, 5, 152360), 'annotation_class_id_1': 27, 'param_1': 1}]
(Background on this error at: https://sqlalche.me/e/20/4xp6)
```

Here's the problematic method:

https://github.com/jacksonjacobs1/QuickAnnotator/blob/4ca883f872bd796e02c8561c64b8424f5a08aeaf/quickannotator/db/crud/tile.py#L331

```python
@staticmethod
@abstractmethod
def get_workers_tiles(annotation_class_id: int, boost_count: int, dialect: Dialects) -> db_models.Tile:
    subquery = (
        select(db_models.Tile.id)
        .where(db_models.Tile.annotation_class_id == annotation_class_id,
            db_models.Tile.gt_datetime.isnot(None))
        .order_by(db_models.Tile.gt_counter.asc(), db_models.Tile.gt_datetime.asc())  # Prioritize under-used, then newest
        .limit(1)
    )

    if dialect == Dialects.POSTGRESQL:
        subquery = subquery.with_for_update(skip_locked=True)
        

    tile = db_session.execute(
        update(db_models.Tile)
        .where(db_models.Tile.id == subquery.scalar_subquery())
        .values(gt_counter=(
                # Only increment selection_count if it's less than max_selections
                case(
                    (db_models.Tile.gt_counter < boost_count, db_models.Tile.gt_counter + 1),
                    else_=db_models.Tile.gt_counter
                )
            ),
            gt_datetime=datetime.now())
        .returning(db_models.Tile)
    ).scalar()

    if tile:
        db_session.expunge(tile)
        return tile
    else:
        return
```

This seems to be resolved by setting `skip_locked=False`, which makes each worker wait for locked rows to unlock.

However, after making this change I receive an error from the logger:
```
    self.connection._commit_impl()
  File "/home/ray/anaconda3/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1147, in _commit_impl
    self._handle_dbapi_exception(e, None, None, None, None)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 2363, in _handle_dbapi_exception
    raise sqlalchemy_exception.with_traceback(exc_info[2]) from e
  File "/home/ray/anaconda3/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1145, in _commit_impl
    self.engine.dialect.do_commit(self.connection)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 715, in do_commit
    dbapi_connection.commit()
sqlalchemy.exc.DatabaseError: (psycopg2.DatabaseError) error with status PGRES_TUPLES_OK and no message from the libpq
(Background on this error at: https://sqlalche.me/e/20/4xp6)
```

This is unrelated to the above error, and I've seen it before. This is the same error I tried to mitigate by calling engine.dispose(close=False) https://docs.sqlalchemy.org/en/20/core/pooling.html#using-connection-pools-with-multiprocessing-or-os-fork. Seems like that fix did not work.

choosehappy: hmm...this seems like a significant issue to resolve? perhaps one can avoid it by using only a single worker, but its not clear to me how this same issue wouldn't crop up if there are multiple distiributed ray acts doing the same process? from the database point of view, this would essentially be the same as multiple threads trying to get a write lock on the rows? is that correct? seems like it would also have a large impact on e.g., patchsorter which would indeed be more focused on having more workers doing more stuff in parallel to keep up the throughput? would a similar function like this be needed there?

jacksonjacobs1: Sorry, didn't get the chance to update this thread.

I discovered that the root issue was that `engine.dispose(close=False)` was being called in the wrong place (the dataset's `__init__`, which get's called once in the parent process). The appropriate place is to call this in the Dataloader's worker_init_fn, which gets called once at the start of each worker fork:

```python
    dataloader = DataLoader(
        patched_dataset,
        batch_size=batch_size_train,
        shuffle=False,  # IterableDataset doesn't support shuffle
        num_workers=num_workers,
        worker_init_fn=lambda worker_id: dispose_engine()  # Dispose engine in each worker to prevent connection issues
    )
```

This is the correct pattern and prevents the postgres error for fresh sessions, but in places where functions use the global db_session (such as the above `get_workers_tiles` method) the session management pattern that I've used is not fork safe, resulting in the same PGRES_TUPLES_OK error shown above. The error means that the session variable points to a non-existent session.

If we want to make it fork safe, I would suggest having all of the ResourceStore classes (AnnotationStore, TileStore, LogStore etc.) accept an optional session argument. If a session object is not supplied, fall back on the global scoped db_session. This would be an appropriate refactor for an opencode agent. I chatted with Opus for a bit, which offered a pretty reasonable implementation plan.

A simpler approach would be to configure the dataloader to use `spawn()` rather than `fork()`. With spawn, the global db_session gets called once per process. Note that ray uses spawn, whereas pytorch dataloader uses fork by default on unix:
https://docs.pytorch.org/docs/stable/data.html#platform-specific-behaviors

What do you think?

choosehappy: i think i understand what you're saying -- but is a scoped session more appropriate? https://docs.sqlalchemy.org/en/21/orm/contextual.html  i think it may do what you're describing directly?

jacksonjacobs1: We use scoped_session to enforce thread safety (necessary for the flask application which is multi-threaded), but this breaks down when a new process is forked. The scoped_session holds a registry across threads, but not processes.

Another thing I found is that ray considers multiprocessing.fork() an anti-pattern and recommends spawn instead. The problems discussed in the guidance don't directly apply to our situation but are good to be aware of:

https://docs.ray.io/en/latest/ray-core/patterns/fork-new-processes.html

choosehappy: got it, this makes sense -- makes me wonder if we should then just use spawn everyone to keep consistent, which would also address this issue, or did i misunderstand? i remember back in the day in pytorch trying spawn vs fork and never really noticated a meaningful difference, not sure if thats changed

jacksonjacobs1: Just made this change:
```python
    dataloader = DataLoader(
        patched_dataset,
        batch_size=batch_size_train,
        shuffle=False,  # IterableDataset doesn't support shuffle
        num_workers=num_workers,
        worker_init_fn=lambda worker_id: dispose_engine(),  # Dispose engine in each worker to prevent connection issues
        multiprocessing_context="spawn"  # Use spawn to avoid issues with the global scoped_session.
    )
```

There was a little additional work (2 lines of code used unpickleable patterns causing problems during the spawn op), but the dataloader seems to be working smoothly now. I've been running the QA instance for about 15 minutes with no database session errors. Nice!

Side note - this DL model has converged pretty nicely on just 64 ground truth annotations. I think Tilak will like this!

image

choosehappy: sssssssssssssswwweet!

PR #133, opened 2026-04-08
Display aggregate counts

jacksonjacobs1:
Display aggregate class-level ground truth counts in the Image table:
image

And in the Projects table, display ground truth counts aggregated over all classes:
image

Counts are queried using a single endpoint call with groupby functionality. The endpoint does not block the loading of images. While pending, count cells display spinners:
image

### DIFF
--- a/quickannotator/api/v1/annotation/routes.py
+++ b/quickannotator/api/v1/annotation/routes.py
@@ -37,15 +37,15 @@ class AnnotationCounts(MethodView):
     def get(self, args):
         """Get annotation counts filtered by project, image, and/or annotation class.
         
-        Use group_by=project to get totals aggregated per project (summed across all images and classes).
+        Use group_by_project=true to get totals aggregated per project (summed across all images and classes).
         """
         project_id = args.get('project_id')
         image_id = args.get('image_id')
         annotation_class_id = args.get('annotation_class_id')
-        group_by = args.get('group_by')
+        group_by_project = args.get('group_by_project', False)
 
         # Determine which projects to iterate
-        if group_by == 'project':
+        if group_by_project:
             if project_id is not None:
                 projects = [db_models.Project.query.get(project_id)]
                 projects = [p for p in projects if p is not None]
--- a/quickannotator/api/v1/annotation/routes.py
+++ b/quickannotator/api/v1/annotation/routes.py
@@ -3,7 +3,6 @@ from quickannotator.constants import PolygonOperations
 from quickannotator.db.crud.annotation import AnnotationStore, build_export_filepath, get_annotation_count
 from quickannotator.db.crud.image import get_image_by_id, get_images_by_project_id
 from quickannotator.db.crud.annotation_class import get_all_annotation_classes_for_project, get_annotation_class_by_id
-from quickannotator.db.crud.project import get_all_projects
 from quickannotator.db.crud.tile import TileStoreFactory
 from quickannotator.db.fsmanager import fsmanager
 from .utils import AnnotationExporter, compute_actor_name, GeometryOperation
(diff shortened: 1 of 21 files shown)


jacksonjacobs1:
Also added similar count endpoints for annotation class and image tables. Count info for these resources displays on the landing page:
image

quickannotator/api/v1/annotation/routes.py
```diff
@@ -28,6 +30,81 @@
...
+        
+        Use group_by=project to get totals aggregated per project (summed across all images and classes).
+        """
+        project_id = args.get('project_id')
+        image_id = args.get('image_id')
+        annotation_class_id = args.get('annotation_class_id')
+        group_by = args.get('group_by')
+
+        # Determine which projects to iterate
+        if group_by == 'project':
```
choosehappy: it looks like group_by is only used here to determine if it is at the project level or not. is that correct? if yes, this is a less favorable pattern. one should avoid string comparisons (prone to e.g., typos and ambiguity (i.e., PROJECT vs project)) and string comparisons are slow. this should instead be a binary, group_by_project = True/False, which is faster and has better tooling surrounding it (e.g., can send true / True / TRUE /1 and the validators should all work no problem)

choosehappy: this is similar to how we swapped a number of other things for enums.

quickannotator/api/v1/annotation/routes.py
```diff
@@ -28,6 +30,81 @@
...
+                        total_gt += get_annotation_count(img.id, ann_cls.id, is_gt=True)
+                        total_pred += get_annotation_count(img.id, ann_cls.id, is_gt=False)
+                results.append({
+                    'project_id': proj.id,
+                    'gt_count': total_gt,
+                    'pred_count': total_pred,
+                })
+            return results, 200
+
+        # Per-image granularity (existing behavior)
```
choosehappy: there are a lot of return statements here where is an anti-pattern. if you look at the structure of this function, it looks this way because its not appropriately following REST - there should be other end points

/counts/image/{id}
/count/images
/count/projects

etc as needed

quickannotator/api/v1/annotation_class/routes.py
```diff
@@ -88,6 +89,29 @@ def get(self, args):
...
+        """Get annotation class counts, optionally grouped by project."""
+        project_id = args.get('project_id')
+        group_by = args.get('group_by')
+
+        if group_by == 'project':
+            if project_id is not None:
+                projects = [p for p in [db_models.AnnotationClass.query.session.get(db_models.Project, project_id)] if p is not None]
+            else:
+                projects = get_all_projects()
+            return [{'project_id': proj.id, 'annotation_class_count': len(get_all_annotation_classes_for_project(proj.id))} for proj in projects], 200
```
choosehappy: similar - this looks like it would be 4 indiviudal endpoints

quickannotator/api/v1/image/routes.py
```diff
@@ -61,6 +63,28 @@ def delete(self, args):
...
+@bp.route('/counts')
+class ImageCounts(MethodView):
+    @bp.arguments(server_models.GetImageCountsArgsSchema, location='query')
+    @bp.response(200, server_models.ImageCountRespSchema(many=True))
+    def get(self, args):
+        """Get image counts, optionally grouped by project."""
+        project_id = args.get('project_id')
+        group_by = args.get('group_by')
+
+        if group_by == 'project':
```
choosehappy: this one looks better since its always returning projects ( the previous one also had counts for images if i remember correctly?)

however, since group_by is _only_ for projects, it should still be a binary. as well, for more clarity, this could still be moved to /counts/images

"counts" should never be "naked" - its an action not a resource

in fact, the correct way to do this is to invert:

/images/counts
/annotations/counts?group_by   

whatever -- so you know what counts your getting back based on the resource in front of it

choosehappy (changes requested):
a few points to consider - please read all the comments - my mind shifted as i was writing the last one and won't go back to update the others so you can see the thought process that got me there

essentially you shouldn't have "/counts" but should instaed have e.g. ,  /annotations/counts  or /images/counts

which accepts  ?group_by_project=True

quickannotator/api/v1/image/routes.py
jacksonjacobs1: I think there may be some confusion here. the prefix for all endpoints is determined by the blueprint. this endpoint for instance is inside the "image" blueprint, so the full path is /image/counts.

Note that /class/counts only returns annotation class counts and /annotation/counts only returns annotation counts.

The approach is to provide 1 /count endpoint per resource, with query params to apply filters and grouping. This may not be strictly restful, but it seems to be a common pattern. Do you agree?

quickannotator/api/v1/annotation/routes.py
jacksonjacobs1: Good point, done.

quickannotator/api/v1/image/routes.py
choosehappy: ah yes, thats right, apologies. in the github review diff i can't see what the bp fixed prefix is.  how you mentioend it is correct - the idea is that /annotations/counts returns all annotations, and then using the query parameters you can _filter_ that set. however, if the query parameters resulted in _different_ data than the original superset, then other end points would be appropriate. e.g.,

/annotation
a1
a2
a3

/annotation?imgid=1
a1
a3

however if you did
/annotation?proj_id=1
and received as feedback
p1
p2
p3

(of different types), this would be incorrect

the first route in this review (/annotations) threw me off because there is a lot of stuff going on in there - it looks like there are some unexpected behaviors possible, e.g., if someone sets both project id and image id, as i undrestand the code it would ignore the latter and only process the former? this usually suggests that the route is "too wide" and should be narrowed  to avoid the ambiguity and reduce potential for misuse?

jacksonjacobs1: Got it. I actually agree that the endpoint for returning annotation counts should change. What about this:

`/project/<project_id>/count/annotations/?image_id,annotation_class_id`

1. no need to worry about grouping, just call the endpoint once per project
2. client can sum counts over image_id / annotation_class_id if needed
3. worst case, add another endpoint for grouped results like `/project/<project_id>/count/annotations/group_by_image/?image_id,annotation_class_id`

The other endpoints should be

`/project/<project_id>/count/annotationclasses` for counting annotationclasses within a project

and 

`/project/<project_id>/count/images` for counting images within a project

... where the root resource is `project` because we are computing a project-level metric (with optional filtering), rather than returning a different type of resource.

choosehappy: i think verbs go _after_ resources

so the more normal way is:

/project/<project_id>/annotationclasses/count/


type / type / verb

instead of type / verb/ type

similar with the others

note as well that the groupby is a filtering and not itself a verb, so this seems more appropriate:

/project/<project_id>/annotations/count/?image_id,annotation_class_id&group_by=image


but its not obvious to me what image_id,annotation_class_id are? are those groups or fields requested?

would it be columns=image_id,annotation_class_id

explicit versus implicit?

jacksonjacobs1: image_id and annotation_class_id are filters. Since it's expensive to perform COUNT on every table, these options allow for a REST call to filter the tables that are COUNTed. They are query params to allow for multiple arguments.

jacksonjacobs1: maybe it would be more correct to accept pairs (image_id1, annotation_class_id1), (image_id1, annotation_class_id2), etc.

jacksonjacobs1: Disregard the last comment - it's too fine-grained for a GET endpoint.

Okay, I'm going to make the following endpoints.

```
/project/<project_id>/annotations/stats/?group_by=<image|annotation_class>&annotation_class_ids=<integer list>&image_ids=<integer list>
```
```json
{
  [
      { "group_id": 12, "group_label": "lymphocyte", "count": 304 },
      { "group_id": 7,  "group_label": "epithelium", "count": 198 }
  ]
}
```

```
/project/<project_id>/annotation_class/stats
```

```json
{
   "count": 3
}
```

```
/project/<project_id>/image/stats
```

```json
{
   "count": 105
}
```

choosehappy: seems correct to me! only small caveat - is stat going to return anything _except_ a count? is there future functioanlity intended to be implemented here? otherwise its a bit strange if "stats" -- returns "count" as the key. it would make more sense that "/count" returns "{"count":xx}" - so its more self documenting and obvious to an end user?

jacksonjacobs1: Yeah, I'm leaving the door open for other stats to be added in the future without having to create new endpoints.

I guess the more correct return structure would be 

```json
{
    "stats": {
           "count": 123
     }
}

```

jacksonjacobs1:
Finished:
image

All stats endpoints return json objects with "count" nested within a "stats" object:
```json
[
  {
    "group_id": 0,
    "group_label": "string",
    "stats": {
      "count": 0
    }
  }
]
```

The client has been updated to reflect this.

PR #134, opened 2026-04-08
Autogenerate tissue mask

jacksonjacobs1:
> Note: do not merge in until #133  is merged. This code reuses one of the endpoints from that branch.

Previously quickannotator did not have a smooth UX when no tissue mask existed for the slide. Here's the new behavior:

- When the user is on the Tissue Mask class and clicks another class, the system automatically checks whether tissue mask annotations exist for the current image via the annotation counts API.
- If a tissue mask already exists, the class switch proceeds automatically with no prompt.
- If no tissue mask exists, the user is blocked from switching and shown a modal dialog:
image

- The modal offers two options:
  - Add my own tissue mask — dismisses the dialog and keeps the user on the Tissue Mask class to draw manually.
  - Auto-generate tissue mask — calls a backend endpoint that generates tissue mask polygons from a downsampled thumbnail using grayscale thresholding and morphological cleanup, then saves them to the database. A spinner is shown while generating.
- Once auto-generation completes, the viewport automatically refreshes to display the generated tissue mask polygons.

> Note: because mask generation completes in under 3 seconds, the client displays a spinner while the mask is generated syncronously by the server. If mask generation had taken longer, an asyncronous ray actor would be spun up.

The tissue mask generation code was inspired by HistoQC / BiopTort
image


quickannotator/api/v1/annotation/models.py
```diff
@@ -50,6 +50,19 @@ class OperationArgsSchema(AnnRespSchema):
     operation = fields.Integer(required=True)  # Default 0 for union.
     polygon2 = db_models.GeometryField(required=True)    # The second polygon
 
+class GetAnnCountsArgsSchema(Schema):
+    project_id = fields.Int(required=False)
+    image_id = fields.Int(required=False)
+    annotation_class_id = fields.Int(required=False)
+    group_by = fields.Str(required=False)  # 'project' to aggregate per-project
```
choosehappy: these were reviewed in another pull request

quickannotator/api/v1/annotation/routes.py
```diff
@@ -1,10 +1,12 @@
 from itertools import product
 from quickannotator.constants import PolygonOperations
-from quickannotator.db.crud.annotation import AnnotationStore, build_export_filepath
-from quickannotator.db.crud.image import get_image_by_id
+from quickannotator.db.crud.annotation import AnnotationStore, build_export_filepath, get_annotation_count
+from quickannotator.db.crud.image import get_image_by_id, get_images_by_project_id
+from quickannotator.db.crud.annotation_class import get_all_annotation_classes_for_project, get_annotation_class_by_id
```
choosehappy, line 5: these were reviewed in another review request

quickannotator/api/v1/annotation/routes.py
```diff
@@ -248,4 +325,30 @@ def generate():
                 while chunk := f.read(constants.STREAMING_CHUNK_SIZE):
                     yield chunk
 
-        return Response(generate(), headers=headers, direct_passthrough=True)
\ No newline at end of file
+        return Response(generate(), headers=headers, direct_passthrough=True)
+
+
+@bp.route('/<int:image_id>/mask/generate')
```
choosehappy, line 256: is this a correct route? i can't see from this part in the code what the bp prefix is -- is this /image/<image_id>?

quickannotator/api/v1/annotation/routes.py
```diff
@@ -248,4 +325,30 @@ def generate():
...
+class GenerateMask(MethodView):
+    @bp.response(200)
+    def post(self, image_id):
```
choosehappy, line 259: not obvious to me what happens here if there is already a tissue mask generated? does this merge in or overwrite?

quickannotator/api/v1/annotation/utils.py
```diff
@@ -320,3 +320,102 @@ def difference(poly1: Polygon, poly2: Polygon, multipoly_func=constants.MultiPol
...
+    Args:
+        image_id: The database ID of the image.
+        disk_size: Radius of the disk structuring element for the minimum filter.
+        threshold: Grayscale intensity threshold (pixels below this are tissue).
+
+    Returns:
+        A list of Shapely Polygon objects in base image coordinate space.
+    """
+    import numpy as np
+    import cv2
```
choosehappy: are all these dependencies in the requirements.txt?

quickannotator/api/v1/annotation/utils.py
```diff
@@ -320,3 +320,102 @@ def difference(poly1: Polygon, poly2: Polygon, multipoly_func=constants.MultiPol
...
+
+    image = get_image_by_id(image_id)
+    if image is None:
+        raise ValueError(f"Image with id {image_id} not found")
+
+    full_path = fsmanager.nas_read.relative_to_global(image.path)
+    slide = large_image.open(full_path)
+
+    # 1. Get a small downsample of the image
+    thumbnail_width = 2048
```
choosehappy: how did you choose this size?

note that the openslide will load the next largest level and downsample in order to do this:

https://github.com/openslide/openslide-python/blob/ff9ef51500897b88a775d37e7b792a8220f3c1d8/openslide/__init__.py#L172


which generally seeems okay - this is a TCGA svs: 

Image

so it would load the 3k x 3k and down sample - however, if the closest level was like 10k x 10k , this would become a bit  more expensive. that said, my intuition is that this would never actually crash - the largest pulled level based on 2k thumbnail would still probably fit in memory- just may take longer than expected to complete. if this is a genuine worry, we can reduce the thumbnail size here to probably 1k without too much of an impact on the downstream results

quickannotator/api/v1/annotation/utils.py
```diff
@@ -320,3 +320,102 @@ def difference(poly1: Polygon, poly2: Polygon, multipoly_func=constants.MultiPol
...
+    thumb_data, _ = slide.getThumbnail(width=thumbnail_width, format='numpy')
+    if thumb_data.shape[2] == 4:
+        thumb_data = thumb_data[:, :, :3]  # drop alpha channel
+
+    thumb_h, thumb_w = thumb_data.shape[:2]
+    scale_x = image.base_width / thumb_w
```
choosehappy, line 361: this is probobably okay for our use case, but note that this could potentially result in a "sub pixel" scaling factor which is a bit strange. the "correct" way to do this, is to find the closest thumbnail size that is a divisor of the base image, so that there is only an integer scaling factor. minor detail

quickannotator/api/v1/annotation/utils.py
```diff
@@ -320,3 +320,102 @@ def difference(poly1: Polygon, poly2: Polygon, multipoly_func=constants.MultiPol
...
+    binary_mask = binary_fill_holes(binary_mask)
+
+    # 4. Convert mask to polygons using OpenCV contours
+    mask_uint8 = binary_mask.astype(np.uint8) * 255
+    contours, hierarchy = cv2.findContours(mask_uint8, cv2.RETR_CCOMP, cv2.CHAIN_APPROX_SIMPLE)
+
+    if not contours:
+        return []
+
+    # Build polygons from contours, handling holes via hierarchy
```
choosehappy: i feel like similar code exists somewhere else in this project and could be reused? e.g., when we generate DL output and convert the predictions to annotations - is that the same process as this?

quickannotator/api/v1/annotation_class/models.py
```diff
@@ -24,3 +24,11 @@ class PutAnnClassArgsSchema(GetAnnClassArgsSchema):
 class SearchAnnClassArgsSchema(Schema):
     name = fields.Str(required=False)
     project_id = fields.Int(required=False)
+
+class GetAnnClassCountsArgsSchema(Schema):
+    project_id = fields.Int(required=False)
+    group_by = fields.Str(required=False)
```
choosehappy: these were reviewed in another review request

choosehappy (changes requested):
some of these code diffs overlapped with another review request - i only focused on the unique ones here. UI looks great -  a few thoughts regarding the backend

quickannotator/api/v1/annotation/routes.py
jacksonjacobs1: This bp has the prefix /annotation. do you think /image is more appropriate?

choosehappy: it seems so, right? the mask is being generated on an image not an annotation

so  /image/1/mask/generate  seems more obvious to read than /anntoation/2/mask/generate

since the "2" should be an annotation ID and not an image id

i.e., GET /annotation/<annotation_id>  is the correct format to get an annotation?

one can then use the /annotation route to get the mask

quickannotator/api/v1/annotation/routes.py
jacksonjacobs1: It will add a duplicate set of tissue mask annotations - there's no flag on the image saying "tissue mask already  generated" 

Basically this is left up to the user, (though the UI won't bring up the "generate mask" popup unless there are no existing tissue mask annotations).

choosehappy: ooo that seems dangerous. if you call it 10 times, you end up with 10 of the exact same polygons on top of each other? 

i would add a parameter here to explicitly let the user control this - by default it should probably be "replace" -- so the parameter would be "append=True" to have the current functionality, but the default would be False (which would replace entirely)

quickannotator/api/v1/annotation/utils.py
jacksonjacobs1: Done. I've copied over the much simpler code in DL prediction - the main difference is that it doesn't allow for holes. K.I.S.S

quickannotator/api/v1/annotation/utils.py
jacksonjacobs1: Note that after the latest commit, the scale_x and scale_y are only applied to polygon coordinates, which are float. So there should be no issue wrt. subpixel scaling.

jacksonjacobs1 (commented):
I think I addressed your comments here. Will respond to comments in #133

quickannotator/api/v1/annotation/utils.py
choosehappy: pew pew pew - if its good enough for the skunkworks ;-)

quickannotator/api/v1/annotation/utils.py
choosehappy: not sure if you saw this one/

quickannotator/api/v1/annotation/utils.py
choosehappy: yes thats right - but even having polyon points which are subpixel is weird - i think the way the current backend works is that when it draws contours on the binary image coming from inference, the points themselves are always pixels, and not subpixels? so all other polygons in that table may be integers except for the mask polygons which would be float? its not a huge deal breaker - the tissue mask one is naturally more coarse. your better argument for leaving it as it is, is that in my suggestion the thumbnails may be of (At least slightly) different magnifications making identfying parameters (e.g., these 5000, 500, below) to work robustly across them more difficult than it needs to be. its easier to push everything into a fixed space size and build an algo there ;-)

choosehappy (approved):
Looks great, nice job!

quickannotator/api/v1/annotation/utils.py
jacksonjacobs1: All polygons saved into the database are float

I made this design choice to match digital slide archive / histomics UI

In retrospect int is probably more desirable than float for our DL-focused application, but I'm not sure it's worth  changing this now?

quickannotator/api/v1/annotation/utils.py
jacksonjacobs1: Okay. I'll change the thumbnail width to 1024

quickannotator/api/v1/annotation/utils.py
choosehappy: nah, that ship has sailed. whats done is done

PR #138, opened 2026-05-11
Dl hardening

### LINKED ISSUE #137: [BUG] infinite/NaN loss errors arising during model training
Ray train workers repeatedly fail, producing this error trace:

```bash
(RayTrainWorker pid=4128523)   return F.conv2d(input, weight, bias, self.stride,
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'seg_bce_pos' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'seg_dice' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'seg_bce_bg' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'edge' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'interior_fill' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'hv' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'recon' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'obj_emb' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'obj_view_cont' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'total_var' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'small_hole' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'consistency' is invalid. Setting to zero.
(RayTrainWorker pid=4128523) [NaN/Inf detected] Loss 'total' is invalid. Setting to zero.
2026-05-08 13:55:42,217 ERROR worker.py:429 -- Unhandled error (suppress with 'RAY_IGNORE_UNHANDLED_ERRORS=1'): r
ay.exceptions.RayTaskError(AssertionError): ray::_Inner.train() (pid=4128320, ip=172.23.0.4, actor_id=d31815c3794
e15bc31c5185407000000, repr=TorchTrainer)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/ray/train/_internal/utils.py", line 57, in check_for_fai
lure
    ray.get(object_ref)
ray.exceptions.RayTaskError(AssertionError): ray::_RayTrainWorker__execute.get_next() (pid=4128523, ip=172.23.0.4
, actor_id=3ccd356676116a7f89c7424907000000, repr=<ray.train._internal.worker_group.RayTrainWorker object at 0x74
bbfe539870>)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/ray/train/_internal/worker_group.py", line 33, in __exec
ute
    raise skipped from exception_cause(skipped)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/ray/train/_internal/utils.py", line 176, in discard_retu
rn_wrapper
    train_func(*args, **kwargs)
  File "/opt/QuickAnnotator/quickannotator/dl/training.py", line 284, in train_pred_loop
    scaler.step(optimizer)
  File "/home/ray/anaconda3/lib/python3.10/site-packages/torch/amp/grad_scaler.py", line 450, in step
    len(optimizer_state["found_inf_per_device"]) > 0
AssertionError: No inf checks were recorded for this optimizer.

The above exception was the direct cause of the following exception:

ray::DLActor.start_dlproc() (pid=4068142, ip=172.23.0.4, actor_id=47141f92b84efd5c9aaaf4fd07000000, repr=<quickan
notator.dl.ray_jackson.DLActor object at 0x74cc61ecd000>)
  File "/opt/QuickAnnotator/quickannotator/dl/ray_jackson.py", line 81, in start_dlproc
    result = trainer.fit()
  File "/home/ray/anaconda3/lib/python3.10/site-packages/ray/train/base_trainer.py", line 722, in fit
    raise TrainingFailedError(
ray.train.base_trainer.TrainingFailedError: The Ray Train run failed. Please inspect the previous error messages 
for a cause. After fixing the issue (assuming that the error is not caused by your own application logic, but rat
her an error such as OOM), you can restart the run from scratch or continue this run.
To continue this run, you can use: `trainer = TorchTrainer.restore("/home/ray/ray_results/TorchTrainer_2026-05-08
_13-55-08")`.
```

Note that all of the `[NaN/Inf detected]` logs occurred at once, so the problem did not originate with a single loss function becoming numerically unstable (which `safe_loss()` would usually handle).

According to ChatGPT, the following flow might be occurring:
```
1. Some layer output becomes NaN/Inf
2. Every downstream loss becomes NaN
3. safe_loss() replaces them with detached zeros
4. Final loss becomes detached
5. AMP scaler sees no inf checks
6. Assertion triggers
```

The implication is that the model has entered a numerically unstable state, but I was unable to reproduce this in live time as the problematic model was loaded from a previous checkpoint.

jacksonjacobs1:
QuickAnnotator will now save the last N model checkpoints for each class, and will automatically load the latest checkpoint.

Additionally, we ensure that all loss output is wrapped in safe_loss to prevent model corruption due to NaN/inf values.


quickannotator/db/fsmanager.py
```diff
@@ -245,6 +250,44 @@ def get_debug_path(self, relative: bool = False):
...
+        Returns the path to the latest model checkpoint for the given annotation class ID, or None if no checkpoint exists.
+        """
+        savepath = self.get_class_checkpoint_path(annotation_class_id)
+        if not os.path.exists(savepath):
+            return None
+        checkpoint_files = [os.path.basename(f) for f in glob.glob(os.path.join(savepath, f"*{constants.CHECKPOINT_FILENAME}"))]
+        if not checkpoint_files:
+            return None
+        latest_checkpoint = max(checkpoint_files, key=lambda f: os.path.getctime(os.path.join(savepath, f)))
+        return os.path.join(savepath, latest_checkpoint)
```
choosehappy, line 264: i see here you're using ctime to define the ordering of files. this feels more like an implicit definition than an explicit definition. generally, if e.g., things get restored from a backup, they may all get the same ctime, and then this measure becomes chaotic. however, you are also the one writing the filename, which allows for an explicit ordering - why not name the files in a way such that the explicit filename results in the correct ordering such as prefix = [yyyy,mm,dd,hh,mm], which should then sort organically correctly?

quickannotator/db/fsmanager.py
```diff
@@ -245,6 +250,44 @@ def get_debug_path(self, relative: bool = False):
...
+        return os.path.join(savepath, filename + "_" + constants.CHECKPOINT_FILENAME)
+
+    def truncate_checkpoints(self, annotation_class_id: int, max_checkpoints: int = 5):
+        """
+        Keeps only the latest `max_checkpoints` checkpoints for the given annotation class ID, and deletes older ones.
+        """
+        savepath = self.get_class_checkpoint_path(annotation_class_id)
+        if not os.path.exists(savepath):
+            return
+        checkpoint_files = [os.path.basename(f) for f in glob.glob(os.path.join(savepath, f"*{constants.CHECKPOINT_FILENAME}"))]
```
choosehappy, line 307: this pattern feels a bit clunky - why not use a deque?

it gets populated at startup, and then it maintains itself organically?


```python
from collections import deque
import os

class RollingFileQueue:
    def __init__(self, maxlen=10):
        self.queue = deque(maxlen=maxlen)

    def push(self, filename):
        if len(self.queue) == self.queue.maxlen:
            # The oldest file is about to be evicted
            oldest = self.queue[0]
            os.remove(oldest)
        self.queue.append(filename)

    def pop(self):
        return self.queue.popleft()
```

quickannotator/dl/training.py
```diff
@@ -314,7 +304,7 @@ def _to_scalar(val):
             #print ("losses:\t",loss_total,positive_mask.sum(),positive_loss,unlabeled_loss)
             
             last_save+=1
```
choosehappy, line 311: approach i mentioned would avoid this type of bookkeeping, which can be a bit fragile

choosehappy (changes requested):
overall good, but take a look at these thoughts - a good pattern to use if possible to simplify things and ensure better consistency. does it fit here?

quickannotator/db/fsmanager.py
choosehappy: btw, i realized that ifyou want to do it via strings, the strings self sort, easier way, similar concept

```python
import bisect
import os

class RollingFilenameQueue:
    def __init__(self, maxlen=10):
        self.queue = []
        self.maxlen = maxlen

    def push(self, filename):
        bisect.insort(self.queue, filename)  # insert in lexographic sorted order
        if len(self.queue) > self.maxlen:
            oldest = self.queue.pop(0)       # evict the lexographically smallest (oldest date)
            os.remove(oldest)

    def pop(self):
        return self.queue.pop(0)             # FIFO — take the oldest (smallest) first

    def peek(self):
        return self.queue[0] if self.queue else None

    def __len__(self):
        return len(self.queue)

    def __repr__(self):
        return f"RollingFilenameQueue({self.queue})"
```


Why this works with your filenames
Since your filenames are YYYYMMDDHHMM, lexicographic order is chronological order — so bisect.insort naturally keeps them oldest → newest:
["202501010800", "202501011200", "202501011600"]
      oldest  →  →  →  →  →  →  →  →   newest
When the queue exceeds 10, pop(0) evicts the oldest file and deletes it from disk.

jacksonjacobs1: Definitely agree that sorting should be done by filename rather than ctime.

I'm not against the dequeue structure, but I think it should be a thin client for the file system rather than an in-memory structure. Otherwise it might fall out of sync if the file system changes while the application is still running (e.g., user deletes checkpoints manually)

Thoughts?

choosehappy: thats okay for me as well - just note that it comes with the overhead of having to do a sort everytime a checkpoint is created. that said, if there are only e.g., 5 or 10 checkpoints, the comptuational overhead is trivial , and your preference for direct alignment with reality is prefered.

however, in that case, a deque is overkill : )   instead it should be: save checkpoint -> glob - > sort -> delete all checkpoints checkpoins[max_checkpoints:]

jacksonjacobs1:
Added the changes to sort by filename rather than ctime.

The logic for saving and truncating checkpoints is now contained within a function save_checkpoint()

quickannotator/dl/training.py
```diff
@@ -314,7 +309,7 @@ def _to_scalar(val):
...
-            if last_save>50:
+            if last_save>constants.CHECKPOINT_PERIOD:
```
choosehappy, line 312: this seems fine, but i would ask do you need last_save, or should this instead be an enumerate somewhere and this here be a mod on the period?

choosehappy (approved):
1 minor question

quickannotator/dl/training.py
jacksonjacobs1: not sure if it's an improvement to change the loop structure from while to foreach + enumerate, since saving the checkpoint is a side effect?

"while" fits here because it's an infinite loop which breaks on a condition that is determined by the user. If we used enumerate, we would need to iterate an object indefinitely and add an if statement checking the condition, which is less natural

choosehappy: ah okay, didn't realize this was a while - totally agree with your assessment

PR #145, opened 2026-07-13
Update openslide to 4.x

jacksonjacobs1:
Update openslide library by installng from PPA. Update openslide python client.


jacksonjacobs1:
Currently trying to find some DICOMs to confirm that the update works as expected.

choosehappy:
> Currently trying to find some DICOMs to confirm that the update works as expected.

https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/

jacksonjacobs1:
Thanks. I tested two of the images with openslide 4.0 and both worked (uncluding Mario's image with a split level 0). As expected, large_image extracted the magnification etc. from the openslide metadata. 

QuickAnnotator's TSV upload approach should support DICOM out of the box, but the drag and drop upload will require the following additional changes:
1. Drag and drop UI accepts .zip files
2. Upload endpoint unzips .zip into nas_write
3. The first (any will do) file within the unzipped folder will saved as the `filepath` column in the database table. 

The final location of the dicom WSI in storage will be: `nas_write/projects/PROJECT_ID/images/IMAGE_ID/dicom_folder`. The "delete image" implementation will not need to change since it removes the IMAGE_ID folder.

choosehappy:
note that dicoms are typically not compressed - most users won't have them in a compressed format. they come off the scanner as multiple dcm files, one per layer. how does that impact things?

jacksonjacobs1:
This would be complicated to implement - a dropzone folder upload would by default upload all of the files as separate objects without retaining directory structure.

if the user uploaded multiple dicoms, or a mix of dicom folders and other files, the backend would have to be responsible for sorting through the .dcm files and regrouping them.

I think requiring compressed archives is (1) easier to implement, (2) typical for web-based uploads and downloads, (3) more efficient for network transfers.

choosehappy:
while i agree, unfortunately doesn't meet the use case

note that openslide or any tool generall that i know of can read the zips - so no user will have .zips in their archive 

i.e., if there is a large read-only S3 bucket filled with 1TB of Dicom WSIs (each WSI gets its own sub directory) - what is the workflow for them to use them? 

if another approach can't be imagined to use teh files in their native format, and if they're forced to use TSV, thats fine, but i wouldn't bother to add any .zip style support - its not going to be used

jacksonjacobs1:
That's fair - yes the workflow for S3 would be to create a goofys mount and use the TSV upload approach.

If we really wanted to allow for folder uploads, it may be possible to write a handler for `DirectoryEntry` items which creates a compressed archive on the fly using something like [JSZip](https://stuk.github.io/jszip/):

https://claude.ai/share/7c89b8f7-6035-4c02-a425-b117d2a6d909

jacksonjacobs1:
But the above approach would not allow for streaming, so it would be quite memory intensive.

choosehappy:
i think this has split into two use cases that are being conflated. previously you were talking about a user randomly dropping files, which requires effort to align into the right layers WSIs. this link you just sent however is talking about sending *directories* (which in this case would be a multiple dcms from a single WSI),  which seems like a non-degenerate case? i.e., if you know all the files in the directory belong to a single WSI, why do they have to be zipped at all instead of uploaded individually and places in the same directory?

jacksonjacobs1:
> why do they have to be zipped at all instead of uploaded individually and places in the same directory?

Zipping the dcms on the client side allows us to keep the same openapi spec as before.

If we just upload the individual dcm files, we need to change the openapi spec to add "group id" metadata for each file uploaded (e.g., files uploaded via folder get a group id). This allows the server to determine which files should be grouped together.

I'm not against it - thoughts?

choosehappy:
this seems like the right way to do -- what is the complexity of implementation?

jacksonjacobs1:
1. update api spec
2. add serverside grouping logic
3. add folder upload support to Dropzone.js

Users would need to understand that folder uploads are strictly reserved for dicom uploads, especially because there's no way for us to validate that a folder is a dicom on the client side. DICOM WSIs may not contain files with .dcm suffixes:

https://openslide.org/formats/dicom/
```
OpenSlide will detect a file as DICOM if:

The file can be parsed as a DICOM Part 10 file.
The file does not have a .tif or .tiff filename extension or cannot be parsed as a TIFF file.
The Media Storage SOP Class UID is VL Whole Slide Microscopy Image Storage (1.2.840.10008.5.1.4.1.1.77.1.6)
```

So we should probably add a visible label for each file uploaded via folder in the UI saying something like 
"DICOM: <FOLDERNAME>"

as well as a note like "Drag and drop files or folders into the dropzone. Note that folder uploads will be interpreted as dicom WSIs"

choosehappy:
i'm okay to accept that all folders are dicom - seems like a reasonable simplfication to me. are there any other breaking changes here? because typically we have 1:1 -- 1 file = 1 WSI. here we have the case where a DICOM is uploaded, and there are e.g., 5 .dcm files - in there that go to 1 WSI, so 5:1 - which is a different ratio. does that negatively impact anything else?

jacksonjacobs1:
The only thing I can think of is, we need to decide what path we store for the image `filepath` column.

If we store the dicom folder `nas_write/projects/PROJECT_ID/images/IMAGE_ID/dicom_folder`

... then we'll need to translate this to a filepath lazily (i.e., before opening the image with large_image. This would require another IO operation each time geojs calls the /tile endpoint.

If we store a filepath
`nas_write/projects/PROJECT_ID/images/IMAGE_ID/dicom_folder/img_1.dcm`

Then the image name column ('dicom_folder') will not match the last part of the filepath 'img_1.dcm', but I think this is fine and is probably the better solution.

choosehappy:
i agree that the latter is more elegant and likely the right thing to do - but in practice i don't know what/if anything this will break?

actually - this may not break anything since essentially if we treat the "biggest" dcm as the DICOM file, then openslide will do the other stuff (reading the different layers/files in the same directory).

in that regrard, i think openslide actually makes us agnostic as long as the files are correctly stored in the same directory - we should be able to use the biggest DCM as a reprensentiative for the file and we're all set? so then quite similar to e.g., an SVS or whatever?

jacksonjacobs1:
I think the file chosen is irrelevant - might as well just choose the first one?

choosehappy:
"maybe" - i know that one of the dcms is also the macro images - i haven't tested what happens if you try to open it instead of  one of them that belong in the micro stack. the _safest_ bet is to take the "largest one" by megabytes - this has been tested and worked robustly

jacksonjacobs1:
I've added DICOM support to QuickAnnotator.
- Same upload workflow (would recommend eventually swapping in the session-based upload workflow from patchsorter v2.0, which has better validation and user feedback)
- If an image is known to be dicom (1. folder_name is included in the drag and drop upload, or 2. TSV upload detects dicom metadata in image file), the name field is set to the folder name rather than the file name.

image

image

image

image

jacksonjacobs1:
Here's an example TSV file for importing DICOMs. Note that putting either the dicom folder, or a specific dicom file will result in the DICOM WSI being correctly stored by QuickAnnotator

```csv
filepath
3DHISTEC
CMU/DCM_0.dcm
```

And the respective folder structure within `nas_read`:

(base) root@481be83afcda:/opt/QuickAnnotator/quickannotator/mounts/nas_read# tree

```
.
├── 3DHISTEC
│   ├── 10_0
│   ├── 11_0
│   ├── 12_0
│   ├── 13_0
│   ├── 1_Map
│   ├── 2
│   ├── 3
│   ├── 4_1
│   ├── 4_2
│   ├── 5_0
│   ├── 6_0
│   ├── 7_0
│   ├── 8_0
│   └── 9_0
└── CMU
    ├── DCM_0.dcm
    ├── DCM_1.dcm
    ├── DCM_2.dcm
    ├── DCM_3.dcm
    ├── DCM_4.dcm
    └── DCM_5.dcm
```

If this looks good, I'll add a "DICOM support" section to the QuickAnnotator documentation

@choosehappy thoughts?

choosehappy (approved):

choosehappy (approved):

choosehappy:
love it, looks good!

PR #148, opened 2026-07-27
[#262] Add "comments" column to image list

### LINKED ISSUE #143: [FEATURE] add "comments" column to image list
## Description  
Users have reported difficulty keeping track of
1. Which images within a project have been assigned to separate annotators
2. Which images have been "completed", i.e., all objects have been annotated.

Adding a sortable, editable "comments" column to the image list on the project page would address these difficulties.

The column would sort lexicographically. Comments can be formatted to take advantage of sorting like so:
"User 1, done"
"User 2, ongoing"
"User 3, done \n Notes: ..."

## Implementation
This feature would require adding a new column "comment" to the image database table.

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/262

### DIFF
--- a/quickannotator/api/v1/image/models.py
+++ b/quickannotator/api/v1/image/models.py
@@ -37,5 +37,8 @@ class UploadFileArgsSchema(Schema):
     project_id = fields.Int(required=True)
     folder_name: str | None = fields.String(load_default=None, allow_none=True)
 
+class UpdateImageCommentSchema(Schema):
+    comment = fields.Str(required=True)
+
 class ImageMetadataRespSchema(Schema):
     mpp = fields.Float()
--- a/quickannotator/api/v1/image/models.py
+++ b/quickannotator/api/v1/image/models.py
@@ -1,8 +1,27 @@
 from flask_smorest.fields import Upload
-from marshmallow import fields, Schema
+from marshmallow import fields, Schema, ValidationError, validates
 from marshmallow_sqlalchemy import SQLAlchemyAutoSchema
+from html.parser import HTMLParser
 
 import quickannotator.db.models as db_models
+
+
+class HTMLTagDetector(HTMLParser):
+    def __init__(self):
+        super().__init__()
+        self.has_tags = False
+
+    def handle_starttag(self, tag, attrs):
+        self.has_tags = True
+
+    def handle_endtag(self, tag):
+        self.has_tags = True
+
+
+def contains_html_tags(html_string: str) -> bool:
+    detector = HTMLTagDetector()
+    detector.feed(html_string)
+    return detector.has_tags
 
 class ImageRespSchema(SQLAlchemyAutoSchema):
     """     Image response schema      """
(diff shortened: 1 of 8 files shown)


jacksonjacobs1:
This PR adds an editable, sortable comments column to the image list on the project page. Any unstructured text can be stored as a comment.

image


Migrating an existing QA database to this version will require the following command:
```sql
ALTER TABLE image ADD COLUMN comment TEXT;
```

quickannotator/api/v1/image/routes.py
```diff
@@ -49,19 +44,33 @@ def post(self, args):
...
+    @bp.arguments(server_models.UpdateImageCommentSchema, location='json')
+    @bp.response(200, server_models.ImageRespSchema)
+    def put(self, args, image_id):
+        """     update image comment   """
+
+        image = get_image_by_id(image_id)
+        if not image:
+            abort(404, message="Image not found")
+
+        image.comment = args['comment']
```
choosehappy, line 68: is there some sanity checking done here to prevent e.g., database injection attacks or other wonky stuff?

choosehappy (changes requested):

quickannotator/api/v1/image/routes.py
jacksonjacobs1: There's no risk of database injection attacks since we use the orm to update the comment column.

According to claude, the main risk is actually XSS attacks, which occurs when a client doesn't unwrap html stored in a TEXT database field. The html can include arbitrary scripts which get run by the browser.

Looks like we can strip html tags on the serverside within the marshmallow schema:

```python
import bleach
from marshmallow import Schema, fields, validate, pre_load

class UpdateImageCommentSchema(Schema):
    comment = fields.Str(required=True, validate=validate.Length(max=1000))

    @pre_load
    def sanitize_comment(self, data, **kwargs):
        if 'comment' in data and data['comment'] is not None:
            # Strip all HTML tags, keep only plain text
            data['comment'] = bleach.clean(data['comment'], tags=[], attributes={}, strip=True)
        return data
```

choosehappy: ah cool

seems cautious - but likely the right thing to do ?

jacksonjacobs1: looks like bleach is a deprecated library. There are other libraries for doing html sanitization, but I don't think it's worth adding dependencies for this minor check.

Instead I'll use the html package which is included in the python standard library.

quickannotator/api/v1/image/models.py
```diff
@@ -1,9 +1,28 @@
 from flask_smorest.fields import Upload
-from marshmallow import fields, Schema
+from marshmallow import fields, Schema, ValidationError, validates
 from marshmallow_sqlalchemy import SQLAlchemyAutoSchema
+from html.parser import HTMLParser
 
 import quickannotator.db.models as db_models
 
+
+class HTMLTagDetector(HTMLParser):
```
jacksonjacobs1, line 9: @choosehappy is this adequate?

jacksonjacobs1: https://docs.python.org/3/library/html.parser.html#example-html-parser-application

choosehappy: yes! seems reasonable to me, thanks - fyi: https://stackoverflow.com/questions/6501583/sqlalchemy-sql-injection

PR #150, opened 2026-07-28
[#263] Retain image list state across application pages

### LINKED ISSUE #144: [FEATURE] Retain image list state across application pages
## Description  
The image list on the QA project page may contain hundreds or thousands of items, and users typically scroll and sort the table to identify images requiring annotation or inspection.

Upon re-entering the project page from the annotation or landing pages, users find that the scroll + sort state has been reset.  This state should be raised to the global application scope (or browser cookies) to improve the uesr experience.

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/263


jacksonjacobs1:
The simplest implementation would just add the image table state to the URL parameters. 

The browser's URL history would then be used to recover the image table state when the user presses the browser's forward or back buttons:
image


This avoids the need to store image table state in global variables or browser cookies.

However, it means that the image table state will not be remembered if the user clicks the project name in the navbar. This behavior is undesirable because it may be unexpected to the lay user. 
image


Instead, when the user clicks the project name in the navbar the application can check the most recent URL in the URL history. If the URL corresponds to the project page, it routes to that URL rather than the default project page URL.

choosehappy:
hmmm interesting problem. maybe this is informative - i'm wondering what the expected behavior is when folks have more than 1 window open. say they open 2 windows, and generate two different table states. 

they then open a 3rd window - what would they expect the state to be? if there is some shared state, we would need a cookies based solution - but i actually don't think that is the case. if i open e.g., gmail in 3 windows, it always defaults to the regular page, no matter what the other states are. this suggests to me that states are totally independent

i feel like generally, when you click the project name (e.g., gmail icon, or these icons:

image


it doesn't actually return you to the saved state, but instead drops you back to the default state? are there cases where the actual previous page state is restored? so i'm not 100% sure that its an _unexpected_ behavior? in thinking about it, it may actually be okay?

jacksonjacobs1:
Looks like both github saves the scroll state in an application cache keyed by URL.

The behavior is a little wonky because github only caches the state on certain user actions, but not on scrollEnd.

This approach is also incompatible with the approach which stores page state in URLs, unless the URL cache cleans URL strings such that the application state data is moved from key -> value in the store.

One note on multiple tabs - React applications are single page applications, so mutliple tabs do not share react app state.

choosehappy:
seems wonky - i'm losing the thread here a bit - i don't think this is mission critical - whats the most reasonable solution? does something actually even need to be done?

jacksonjacobs1:
the most reasonable solution is probably to put the state in the URL and call it a day. I'm fine with this

choosehappy:
> the most reasonable solution is probably to put the state in the URL and call it a day. I'm fine with this

oooooookay for me

PR #151, opened 2026-07-28
Feature/264 update export filename format

### LINKED ISSUE #146: [FEATURE] Update export filename format
## Description  
.geojson files exported from QA currently follow the naming format:

`annotation_{IMAGE_ID}_{CLASS_ID}_gt_{DATETIME}_.geojson`

This should be updated to match the file naming format expected for uploads:

`{IMAGE_NAME}_{CLASS_NAME}.geojson`

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/264

### DIFF
--- a/quickannotator/api/v1/image/utils.py
+++ b/quickannotator/api/v1/image/utils.py
@@ -85,7 +85,10 @@ def _import_annotations_from_temp(image_id, file_basename):
         annot_cls_id = annot_cls.id
         for fmt in constants.AnnotationFileFormats:
             temp_path = fsmanager.nas_write.get_temp_path(relative=False)
-            annotation_filename = fsmanager.nas_write.construct_annotation_file_name(file_basename, annot_cls_name, fmt.value)
+            annotation_filename = fsmanager.nas_write.construct_annotation_file_name(
+                image_name=file_basename, 
+                annotation_class_name=annot_cls_name,
+                extension=fmt.value)
             annot_filepath = os.path.join(temp_path, annotation_filename)
             if os.path.exists(annot_filepath):
                 logger.info(f"Found image annotation file - {annot_filepath}")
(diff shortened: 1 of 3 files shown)


choosehappy (approved):

jacksonjacobs1:
Another issue that should be addressed in this PR is the ambiguity of our naming format for geojson files:

`{IMAGE_NAME}_{CLASS_NAME}.geojson`

The `_` leads to ambiguity because this character is extremely common in file names, so IMAGE_NAME (which is derived from a filename) is likely to include one or more underscores. We should choose a different delimeter which is less common in filenames, e.g., a double underscore.

choosehappy:
there should also be sanitizaton of the image and class names

i.e., altough the UI likely should and does support spaces in names, when exporting we should never generate filenames with spaces in them, they should be replaced with "-" or "_" 

this starts to become a bit complex and i'm wondering if we're overthinking this and then will over engineer it. i wonder if there is a more standaraized way of dealing with this? we're likely not the first ones to encounter this issue

jacksonjacobs1:
Looks like this is the method that ships with Flask:
https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html
```python
>>> secure_filename("My cool movie.mov")
'My_cool_movie.mov'
>>> secure_filename("../../../etc/passwd")
'etc_passwd'
>>> secure_filename(u'i contain cool \xfcml\xe4uts.txt')
'i_contain_cool_umlauts.txt'
```

But the implementation should be shared between QA and PS, so id instead opt for something lightweight like pathvalidate:
https://github.com/thombashi/pathvalidate

choosehappy:
sure, fine for me - my question overall was how to balance e.g., user inputs versus QA file formats, versus import/exporting - etc

jacksonjacobs1:
DSA is probably a good reference since that platform also handles drag and drop and server-side file system uploads.

DSA allows spaces in the UI and keeps the spaces in image downloads. Spaces in filenames don't concern me much on the application side - it seems more like something that we leave up to the user to e.g., make sure they correctly handle download URLs containing spaces?

DSA does clean some characters, for instance `/` is replaced with `:` in uploaded filenames.

choosehappy:
okay, i guess that makes sense to me. if the user is doing stuff with spaces, technically we can be accpeting of it and working around - even if it isn't best practice. i think we can follow the DSAs lead here

PR #152, opened 2026-07-28
Feature/266 add ui slider to control prediction threshold

jacksonjacobs1:
https://jacksonjjacobs.com/openproject/work_packages/266

### DIFF
--- a/quickannotator/api/v1/ray/routes.py
+++ b/quickannotator/api/v1/ray/routes.py
@@ -9,6 +9,7 @@ from ray.util.state.common import TaskState
 from ray.util import state
 from ray.util.state.exception import ServerUnavailable
 from quickannotator.db.crud.annotation_class import build_actor_name
+from quickannotator.db.crud.tile import TileStoreFactory
 
 bp = Blueprint('ray', __name__, description='Ray operations')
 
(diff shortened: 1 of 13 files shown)


quickannotator/db/crud/tile.py
```diff
@@ -288,6 +288,18 @@ def reset_all_PROCESSING_tiles(annotation_class_id: int):
         db_session.execute(stmt)
         db_session.commit()
 
+    @staticmethod
+    def reset_all_PROCESSING_and_DONEPROCESSING_tiles(annotation_class_id: int):
```
jacksonjacobs1, line 292: Method for setting all tiles (except those currently processing) back to the UNSEEN state.

quickannotator/dl/ray_jackson.py
```diff
@@ -42,6 +42,7 @@ def __init__(self, annotation_class_id: int, tile_size: int, magnification: floa
         self.allow_pred = True  # --- don't know if we'll ever need this, so hard setting to true
         self.enable_training = True
         self.magnification = magnification
+        self.inference_threshold: float = 0.5
```
jacksonjacobs1, line 45: Inference threshold added as a ray actor field

quickannotator/api/v1/ray/routes.py
```diff
@@ -80,6 +81,44 @@ def post(self, args, annotation_class_id):
...
+        """
+        actor_name = build_actor_name(int(annotation_class_id))
+        try:
+            actor = ray.get_actor(actor_name)
+        except ValueError:
+            return abort(404)
+        except ray.exceptions.GetTimeoutError:
+            return abort(408)
+        
+        ref = actor.set_inference_threshold.remote(args['threshold'])
```
jacksonjacobs1, line 102: set_inference_threshold is called remotely on the actor, going into effect in the next inference batch processed by the actor.

jacksonjacobs1:
I would also like to suggest the following additional DL control changes:
1. Model training should be frozen by default. This will prevent the user from accidentally allowing a model to train when they initially open the annotation page / switch classes.
2. Add a UI button to kill a DL actor. Currently DL actors are only killed when the user switches classes, a mechanism designed to prevent too many DL actors from being active at the same time. However, there's no way to shut down a DL actor short of stopping the QuickAnnotator process that spawned the actor.

jacksonjacobs1:
I'll merge this PR in on monday - the changes suggested above can be implemented in future PRs.

quickannotator/db/crud/tile.py
choosehappy: are you manually adding these comments in to github or is it automated? i guess a question is - would they be better done as comments witin the code itself?

quickannotator/dl/ray_jackson.py
choosehappy: these are all binary models, is that right? because note that if you have a 3 class system, you only need .31 to yield the argmax equivalent - the threshold is essentially 1/N

quickannotator/dl/training.py
```diff
@@ -211,6 +211,7 @@ def train_pred_loop(config):
 
     # procRunningSince will be None if the DL processing is to be stopped, resulting in the model being unloaded from GPU and the training loop exiting
     while ray.get(myactor.get_proc_running_since.remote()):    
+        inference_threshold = ray.get(myactor.get_inference_threshold.remote())
```
choosehappy, line 214: how long does this take? is it slow and/or significantly blocking? one could e.g., make this an async process so that it gets updated "whenever" it gets updated - to at least make sure that the training keeps pushing ahead

choosehappy (commented):
Some comments for consideration - if they're not relevant, then okay to merge!

quickannotator/dl/training.py
jacksonjacobs1: Extremely fast, just a few ms and not a bottleneck.

quickannotator/db/crud/tile.py
jacksonjacobs1: I manually added these github comments to highlight areas for review

quickannotator/dl/ray_jackson.py
jacksonjacobs1: Yep all binary

